PR #3, opened 2019-12-09
Fixes and format handling improvements

bcortier-devolutions:


picky-server/src/main.rs
```diff
@@ -75,3 +54,24 @@ fn init_logs(config: &ServerConfig) {
...
+        include_str!("../../test_assets/private_keys/rsa-2048-pk_4.key");
+    pub const RSA_2048_PK_5: &str =
+        include_str!("../../test_assets/private_keys/rsa-2048-pk_5.key");
+    pub const RSA_2048_PK_6: &str =
+        include_str!("../../test_assets/private_keys/rsa-2048-pk_6.key");
+    pub const RSA_4096_PK_1: &str =
+        include_str!("../../test_assets/private_keys/rsa-4096-pk_1.key");
+    pub const RSA_4096_PK_2: &str =
+        include_str!("../../test_assets/private_keys/rsa-4096-pk_2.key");
+    //pub const RSA_4096_PK_3: &str = include_str!("../../test_assets/private_keys/rsa-4096-pk_3.key");
```
ekse: Is this commented line still needed?

bcortier-devolutions: I would like to uncomment once rsa crate support this key, but it's technically not needed

Issue #14, opened 2020-01-13
Asn1RawDer with Application/Context tags


doshmajhan:
Something I noticed when using the new Asn1RawDer type _(thank you by the way for adding that)_ is if you wrap it in something like `ApplicationTag0` the Application tag doesn't get included when serializing the data.

For example if you have `ApplicationTag0(Asn1RawDer(to_vec(&data).unwrap()))`  it would be seralized into something like:
```
SEQUENCE
   ... etc
```
instead of 
```
[0]
    SEQUENCE
        ... etc
```

Is this the intended functionality with the `Asn1RawDer` type?

bcortier-devolutions:
Thank you for reporting the issue. I'll have a look this week.

bcortier-devolutions:
Fixed in `picky-asn1-der` v0.2.2

PR #15, opened 2020-01-14
Provisioner issued Json Web Token identification for signing requests

bcortier-devolutions:
Also fixes https://github.com/Devolutions/picky-rs/issues/14


picky-server/src/configuration.rs
```diff
@@ -15,6 +15,8 @@ const PICKY_INTERMEDIATE_CERT_ENV: &str = "PICKY_INTERMEDIATE_CERT";
 const PICKY_INTERMEDIATE_KEY_ENV: &str = "PICKY_INTERMEDIATE_KEY";
 const PICKY_SAVE_CERTIFICATE_ENV: &str = "PICKY_SAVE_CERTIFICATE";
 const PICKY_BACKEND_FILE_PATH_ENV: &str = "PICKY_BACKEND_FILE_PATH";
+const PICKY_DEN_PUBLIC_KEY_FILE_ENV: &str = "PICKY_DEN_PUBLIC_KEY_FILE";
```
awakecoding: rename to "PICKY_PUBLIC_KEY_FILE" instead of "PICKY_DEN_PUBLIC_KEY_FILE", same for the private key

PR #20, opened 2020-01-20
Update addressing format as specified in spec doc

bcortier-devolutions:
Migration is automatically performed if v0 schema is detected.


bcortier-devolutions:
@awakecoding Okay, I'll make another PR updating these later then!

PR #22, opened 2020-01-27
HTTP signature parser and builder legacy support

bcortier-devolutions:
- builder do not add spaces between fields of the produced signature
- builder do not add quotes on integers (eg: created=2923983 instead of
created="2923983")
- parser supports values with quotes or no quotes at all regardless of
the parameter

That is, produced HTTP signatures by builder are now like:
```
Signature
keyId="my-rsa-key",created=1402170695,headers="(request-target)
(created) date",signature="..."
```

Parser can read anything like above and even weirdly formatted
signatures like:
```
Signature keyId = my-rsa-key, created=
"1402170695",headers=(request-target) (created) date, signature=...
```

You can build a legacy HTTP signature by calling the `legacy` method on the HTTP Signature builder (see legacy test).

### DIFF
--- a/picky/src/http/http_signature.rs
+++ b/picky/src/http/http_signature.rs
@@ -3,7 +3,7 @@ use crate::{
     key::{PrivateKey, PublicKey},
     signature::{SignatureError, SignatureHashType},
 };
-use base64::DecodeError;
+use base64::{DecodeError, URL_SAFE_NO_PAD};
 use snafu::Snafu;
 use std::{
     borrow::Cow,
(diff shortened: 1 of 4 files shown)


picky/src/http/http_signature.rs
```diff
@@ -229,8 +252,11 @@ impl FromStr for HttpSignature {
         for item in items {
             if let Some(index) = item.find('=') {
                 let (key, value) = item.split_at(index);
-                if value.len() >= 3 {
-                    keys.insert(key.trim(), value[2..value.len() - 1].to_owned());
+                let value = value[1..].trim();
+                if value.starts_with('"') && value.len() >= 2 {
```
fdubois1: The function trim_matches (or trim_start_matches and trim_end_matches) could be used instead of doing an if. But both work.

picky/src/http/http_signature.rs
```diff
@@ -272,6 +298,14 @@ impl FromStr for HttpSignature {
             None
         };
 
+        let signature =
+            keys.remove(HTTP_SIGNATURE_SIGNATURE)
+                .ok_or_else(|| HttpSignatureError::MissingRequiredParameter {
+                    parameter: HTTP_SIGNATURE_SIGNATURE,
+                })?;
+
+        let legacy = !signature.ends_with('=');
```
fdubois1: Could we have case where we don't have to add padding in base64 "normal" mode ? I think it is possible, no ?

bcortier-devolutions: Should never happen with SHA256, but might indeed happen with something else. Let's be safe, I'll check for '/' and '+' characters instead.

picky/src/http/http_signature.rs
bcortier-devolutions: Nice catch.

PR #24, opened 2020-02-04
Make picky-server HTTP controller async with saphir 2

bcortier-devolutions:


picky-server/CHANGELOG.md
```diff
@@ -1,5 +1,27 @@
 # Changelog
 
+## [4.5.0] 2020-03-18
```
fdubois1: I would suggest to keep the date empty for now or write 2020-xx-yy.  We will release it probably later, when it will be needed

PR #30, opened 2020-03-18
Added changelog for picky_server

fdubois1:


picky-server/CHANGELOG.md
```diff
@@ -0,0 +1,8 @@
+# Changelog
+
+## [4.4.0] 2020-03-13
+
+### Fixed
```
bcortier-devolutions: I believe this isn't a fix?

```suggestion
### Changed
```

fdubois1: This is why I prefer the listing format like Marc did in WaykDen-ps ;-)  I don't care. If I remember, it was a fix in mongodb-cwal, this is why I added it here as a fixed... but could be a change because we didn't support it before and now we support it.

bcortier-devolutions: Okay! I guess it could be "Added" in this case.
For reference: https://keepachangelog.com/en/1.0.0/
But as long as we all understand what's going on, I'm okay with it!
Feel free to update or merge. :+1:

PR #32, opened 2020-04-16
Verifier API for Cert + leeway for HTTP signatures

bcortier-devolutions:


picky-server/src/http/authorization.rs
```diff
@@ -123,7 +123,7 @@ mod tests {
 
     fn get_provider_token(private_key: &PrivateKey) -> String {
         let claims = ProviderClaims {
-            x509_duration_secs: None,
+            x509_duration_secs: 7_776_000,
```
fdubois1, line 126: Could you add a comment saying how long is it in days or months ?

picky-server/src/http/controller.rs
```diff
@@ -678,7 +680,7 @@ mod tests {
         )
         .expect("couldn't generate csr");
 
-        let signed_cert = block_on(sign_certificate(&ca_name, csr, &config, storage.as_ref(), None))
+        let signed_cert = block_on(sign_certificate(&ca_name, csr, &config, storage.as_ref(), 7_776_000))
```
fdubois1, line 683: You could use the constant or could add a comment to specify how long it is.

PR #34, opened 2020-05-04
Fix implementations of object identifiers

ionut-arm:
Fixing the implementations of AES and SHA object identifiers by adding
the parameters they need. Also added a few unit tests for them.

Signed-off-by: Ionut Mihalcea <ionut.mihalcea@arm.com>


bcortier-devolutions:
Thank you! LGTM :slightly_smiling_face: 

I re-based your branch and modified only one place: I put tests behind the compiler flag "test" with `#[cfg(test)]`
Also, since this is breaking change (API is changed) I'll merge in "picky-lib-5.0.0" branch and publish as v5.0.0. I intend to do that today :slightly_smiling_face:

PR #39, opened 2020-06-17
picky-server: use official mongo driver instead of prototype

bcortier-devolutions:
I also published a re-usable [crate](https://crates.io/crates/mongodm) for the ORM part ([doc](https://docs.rs/mongodm/0.1.0/mongodm/))


bcortier-devolutions:
I just transferred ownership to `Devolutions` :+1:

PR #41, opened 2020-07-10
No more api key

bcortier-devolutions:
cf: WAYK-2038 ticket in bitbucket.


bcortier-devolutions:
This PR also update picky server changelog to specificy that API key has been removed. Is there other information you see we should add?

fdubois1:
Oh, my bad, everything is fine thanks

PR #61, opened 2020-09-21
Fix invalid AAD for AES-GCM in JWE

bcortier-devolutions:
I'm not sure what is correct way because RFCs seems to not agree, but I changed to match [`go-jose`](https://github.com/square/go-jose) implementation so that picky implementation is compatible with [`step-cli`](https://github.com/smallstep/cli).

### Why implementations were incompatibles?

[RFC 7518](https://tools.ietf.org/html/rfc7518#section-5.3) says

> Content Encryption with AES GCM
> […]
> The Additional Authenticated Data value used is the empty octet string.

But example in [RFC 7516](https://tools.ietf.org/html/rfc7516#appendix-A.1.5) says

> Let the Additional Authenticated Data encryption parameter be ASCII(BASE64URL(UTF8(JWE Protected Header))).

This is how `go-jose` is proceeding as well, so I changed to the same behavior.


awakecoding (approved):
LGTM

PR #69, opened 2020-10-21
Fix serial number generation and RSA's second exponent calculus

### LINKED ISSUE #67: Possible key generation incompatibility with ring
I was working on a code snippet to generate a fresh key pair and hand it to tokio-rustls (which uses [ring](https://crates.io/crates/ring) for its crypto work).  However, the keys generated by PrivateKey::generate_rsa don't seem to make it happy.  If you bypass rustls, as in this test program:
``` rust
fn main() {
    let server_private_key = picky::key::PrivateKey::generate_rsa(2048).unwrap();
    let res = ring::signature::RsaKeyPair::from_pkcs8(&server_private_key.to_pkcs8().unwrap());
    eprintln!("Result = {:?}", res);
}
```
you'll get back `Err(KeyRejected("InconsistentComponents"))`

Unfortunately, `ring` uses this error for *many* conditions, but I tracked it down to this particular check in `limb.rs:parse_big_endian_and_pad_consttime()`:
``` rust
    if num_encoded_limbs > result.len() {
        return Err(error::Unspecified);
    }
```
It fails due to the encoded limbs being 32, but result.len() being 16.

I'm at a bit of a loss and was hoping this might ring a bell (no pun intended) with the picky development team.  It's presumably an issue with either the `rsa` crate (which picky uses) or `ring`, but I'm not positive which direction to head first.

### LINKED ISSUE #68: Negative serial numbers not accepted by RusTLS
picky's certificate builder provides only random serial numbers, with no [obvious] way to override them.  Unfortunately, the generator can produce negative values, which causes tokio-rustls to fail to connect to a server using one of these with "invalid certificate: BadDER".  The issue appears to be in `generate_serial_number()`, which can produce a number that can go negative when stored in a `IntegerAsn1`(a signed type):

```
fn generate_serial_number() -> Vec<u8> {
    let x = rand::random::<u32>();
    let b1 = ((x >> 24) & 0xff) as u8;
    let b2 = ((x >> 16) & 0xff) as u8;
    let b3 = ((x >> 8) & 0xff) as u8;
    let b4 = (x & 0xff) as u8;
    vec![b1, b2, b3, b4]
}
```

I think the quick fix is probably just to mask off `b1` with `0x7f`.

bcortier-devolutions:
Closes #67 
Closes #68

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1685,6 +1685,7 @@ dependencies = [
  "picky-asn1-x509",
  "pretty_assertions",
  "rand",
+ "ring",
  "rsa",
  "serde",
  "serde_json",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1722,7 +1722,7 @@ dependencies = [
 
 [[package]]
 name = "picky-asn1-x509"
-version = "0.3.3"
+version = "0.3.4"
 dependencies = [
  "base64 0.12.3",
  "hex",
--- a/picky-asn1-x509/CHANGELOG.md
+++ b/picky-asn1-x509/CHANGELOG.md
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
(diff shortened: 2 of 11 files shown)


awakecoding (approved):
LGTM

Issue #77, opened 2021-02-22
[Question] How to add "hostname" attribute to CertificationRequestInfo?


Geobert:
Hi,

We first used `Csr::generate` to create a csr but now we'd like to add a hostname in it so I've duplicated the code of generate in order to modify the `CertificationRequestInfo`. But looking at the documentation, I have no clue on how to add an "hostname" attribute and its value.

Thank you for any insight :D

CBenoit:
Hi @Geobert,

I assume you're referring to [CertificateRequestInfo attributes field](https://tools.ietf.org/html/rfc2986#section-4)?
These are not yet supported by picky: [certification_request.rs#L20](https://github.com/Devolutions/picky-rs/blob/master/picky-asn1-x509/src/certification_request.rs#L20)

Geobert:
Thank you for your answer!

> These are not yet supported by picky :/ (see certification_request.rs#L20)

I see, is there any plan to support them and when?

CBenoit:
> Thank you for your answer!
> 
> > These are not yet supported by picky :/ (see certification_request.rs#L20)
> 
> I see, is there any plan to support them and when?

I'm afraid to say there is no plan because no one needs that at @Devolutions AFAIK. However, I would gladly accept and review PR to improve on that :slightly_smiling_face:

Geobert:
>  I would gladly accept and review PR to improve on that 🙂

Hi, thank you for your answer!

What would it take to add support to add attributes to a CSR?

Looking at the code, I'm not sure how to modify the `attributes` field.
My first guess is to change its type for something like `Implicit<Option<Something>>`
`Something` being a type that can store a key and its value, ordered and serializable to DER.

Am I on the right track?

CBenoit:
> > I would gladly accept and review PR to improve on that slightly_smiling_face
> 
> Hi, thank you for your answer!
> 
> What would it take to add support to add attributes to a CSR?
> 
> Looking at the code, I'm not sure how to modify the `attributes` field.
> My first guess is to change its type for something like `Implicit<Option<Something>>`
> `Something` being a type that can store a key and its value, ordered and serializable to DER.
> 
> Am I on the right track?

Exactly, you are on the right track! :+1: 

Looking at the RFC I see this is behind Application Tag 0 (notice [0])

![image](https://user-images.githubusercontent.com/3809077/109026859-33b70c80-768e-11eb-98bd-9a5e127e9d8a.png)

As such this should probably be something like
```
[...]

PR #78, opened 2021-03-02
Add extension request support

Geobert:
Address #77 , on behalf of Isode Ltd.

Some precision on this PR:

# attributes field type

the `attributes` field ended up being `Implicit<Option<ApplicationTag0<Attribute>>>` instead of 
`Implicit<Option<ApplicationTag0<Asn1SetOf<Attribute>>>>` because after testing the output of openssl, it seems that the Asn1SetOf is not present.

using the attached files, you can use
`openssl req -config req.txt -new -key key.pem -out req.pem` 

to generate the CSR and see that there's no `SET` after `[0]`

I was curious on what's the output of openssl if we add a challengePassword to the request: add `attributes = req_attributes` to the `[ req ]` section and rerun the openssl command, you'll see:

image

so after the `[0]`, no `SET` and two consecutive `SEQUENCE`.

We don't need challengePassword yet, but if we add support later, I think `ApplicationTag0` will need some way to have a `SET` without serializing it.

# Unit test

I've added a Unit test for extensions using the config attached and openssl to create the PEM test data.
As attributes has change default value, the `deserialize_csr` test is broken, I think it needs to be updated but I prefer to ask before doing anything on this.
EDIT: maybe it's not the test, but the encoding, I've lost the `[0]` so ignore this point

# Copyright

The legal team of my company is happy to license the contribution under the dual-license MIT-Apache-2.0 but would like to have Isode Ltd. mentioned somewhere. Some crates have a `Contributor.txt` file listing contributors. Would you (and Devolutions) agree to add this file to the repository? It would look like this (header copied from `oxide-auth` crate's repository):
```
# A list of people who have contributed to this project, with optional metadata
# The list is sorted chronologically, excluding the current maintainer who should always be listed first
# Format: Foo Bar <optional mail address> (comment e.g. company affiliation)

Benoît Cortier (current maintainer)
Jonathan Trépanier (on behalf of Devolutions)
Marc-André Moreau (on behalf of Devolutions)
Jean-Francois Theroux
François Dubois
Sébastien Duquette (on behalf of Devolutions)
Johann Dufaud
Richard Markiewicz
Ionuț Mihalcea
Kim Altintop
Joe Ellis
Hugues de Valon
Geobert Quach (on behalf of Isode Ltd.)
```
I took the time to sort the different contributors chronologically :)

# Conclusion and Recap

As the message is quite large, here is a summary on what's left:
- [x] agree on the type of `attributes`
- [ ] fix the unit test `deserialize_csr` because of change of `attributes` type
- [x] agree on adding a contributors list file

[req.zip](https://github.com/Devolutions/picky-rs/files/6069344/req.zip)

### DIFF
--- a/picky-asn1-x509/Cargo.toml
+++ b/picky-asn1-x509/Cargo.toml
@@ -1,7 +1,14 @@
 [package]
 name = "picky-asn1-x509"
 version = "0.4.0"
-authors = ["Benoît CORTIER <benoit.cortier@fried-world.eu>"]
+authors = [
+    "Benoît CORTIER <benoit.cortier@fried-world.eu>",
+    "Sergey Noskov <snoskov@avito.ru>",
+    "Kim Altintop <kim@monadic.xyz>",
+    "Joe Ellis <joe.ellis@arm.com>",
+    "Hugues de Valon <hugues.devalon@arm.com>",
+    "Isode Ltd./Geobert Quach <geobert.quach@isode.com>"
+]
 description = "Provides ASN1 types defined by X.509 related RFCs"
 edition = "2018"
 license = "MIT OR Apache-2.0"
--- a/picky-asn1-x509/src/certification_request.rs
+++ b/picky-asn1-x509/src/certification_request.rs
@@ -44,7 +44,8 @@ impl CertificationRequestInfo {
     }
 }
 
-// this type is a hack to workaround [this issue](https://github.com/Devolutions/picky-rs/pull/78#issuecomment-789904165).
+// FIXME: this type is a hack to workaround [this issue](https://github.com/Devolutions/picky-rs/pull/78#issuecomment-789904165).
+// Further refactorings are required to clean this up (proper support for IMPLICIT / EXPLICIT tags, etc)
 #[derive(Clone, Debug, PartialEq)]
 pub struct Attributes(pub Vec<Attribute>);
 
--- a/picky-asn1-x509/src/certification_request.rs
+++ b/picky-asn1-x509/src/certification_request.rs
@@ -1,6 +1,6 @@
-use crate::{AlgorithmIdentifier, Name, SubjectPublicKeyInfo};
-use picky_asn1::wrapper::{ApplicationTag0, BitStringAsn1, HeaderOnly, Implicit};
-use serde::{Deserialize, Serialize};
+use crate::{oids, AlgorithmIdentifier, Extension, Extensions, Name, SubjectPublicKeyInfo};
+use picky_asn1::wrapper::{ApplicationTag0, Asn1SetOf, BitStringAsn1, Implicit, ObjectIdentifierAsn1};
+use serde::{de, ser, Deserialize, Serialize};
 
 /// [RFC 2986 #4](https://tools.ietf.org/html/rfc2986#section-4)
 ///
@@ -17,8 +17,7 @@ pub struct CertificationRequestInfo {
     pub version: u8,
     pub subject: Name,
     pub subject_public_key_info: SubjectPublicKeyInfo,
-    /// Not yet supported
-    pub attributes: Implicit<Option<HeaderOnly<ApplicationTag0<()>>>>,
+    pub attributes: Implicit<Option<ApplicationTag0<Attribute>>>,
 }
 
 impl CertificationRequestInfo {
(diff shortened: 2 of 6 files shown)


Geobert:
On the test failing, I tried to change the type from `Implicit<Option<ApplicationTag0<Attribute>>>` to `Implicit<ApplicationTag0<Option<Attribute>>>` to try to get back that `[0]` but it seems that I need the `HeaderOnly<ApplicationTag0<()>>` construct for the serialization to work?

CBenoit:
Hi @Geobert , thank you! Great first PR :tada: 

> the attributes field ended up being Implicit<Option<ApplicationTag0<Attribute>>> instead of  Implicit<Option<ApplicationTag0<Asn1SetOf<Attribute>>>> because after testing the output of openssl, it seems that the Asn1SetOf is not present.
> ...
> I was curious on what's the output of openssl if we add a challengePassword to the request: add attributes = req_attributes to the [ req ] section and rerun the openssl command, you'll see:
> …
> so after the [0], no SET and two consecutive SEQUENCE.

Sorry I should have elaborate a bit more on what is an application tag in ASN.1. Here [0] means that "something" is behind the application tag number 0. DER representation is basically a tag (type) followed by a length and some content.
For instance, integer has tag `0x02`. So you'll have something like `0x028181008F…`.
However, let's consider a piece of the definition of the X509 TBSCertificate v3 from RFC 5280:

```
   TBSCertificate  ::=  SEQUENCE  {
        version         [0]  EXPLICIT Version DEFAULT v1,
        serialNumber         CertificateSerialNumber,
        …
        }

 Version  ::=  INTEGER  {  v1(0), v2(1), v3(2)  }

 CertificateSerialNumber  ::=  INTEGER
``` 

As you can see `version` has a default value of v1 (v1 definition had no `version` field). However, both `version` and `serialNumber` are `INTEGER`. How one should determine that the first integer in the `SEQUENCE` is the `version` and not the `serialNumber`? Yes, that's why we have application tags. The application tag is used to disambiguate encoding by _replacing the actual tag with another one_. As such, `version` encoding is not something like `0x028181008F…` but `0xA08181008F…`: `0x02` is replaced by `0xA0` (application tag 0).

That's exactly what is happening here. `asn1js` can't tell that this [0] is actually a `SET` because this information is not replaced by the application tag. Instead you need to rely on the definition in the RFC. That being said, you can see there are multiple elements under the [0] so we can tell this must be either a `SEQUENCE` or a `SET` just by looking at it.

As such `Implicit<Option<ApplicationTag0<Asn1SetOf<Attribute>>>>` is still the right way to do it.

This also solves your issue with the test. The failing test is expecting an _empty_ `SET` (encoded `0xA000`).

By the way I think the `Implicit<Option<…>>` part is not really required: by looking at the RFC again it seems this field is not marked as IMPLICIT so this should always be present. Maybe you can try to simplify that and see if tests are still green (we test against mozilla mkcert).

I asked my team leader @awakecoding about the `Contributor.txt`. We usually simply use the `package.authors` field in `Cargo.toml` for this purpose. I believe that's the standard way to proceed in the Rust ecosystem, and names are displayed on `crates.io`:
image
Given that you contributed to `picky-asn1-x509` crate you would add your name in [this Cargo.toml](https://github.com/Devolutions/picky-rs/blob/master/picky-asn1-x509/Cargo.toml#L4).
Would this be alright for you?
If it's okay for you, you could also add
```
"Sergey Noskov <snoskov@avito.ru>"
"Kim Altintop <kim@monadic.xyz>"
"Joe Ellis <joe.ellis@arm.com>"
"Hugues de Valon <hugues.devalon@arm.com>"
"Geobert Quach <geobert.quach@isode.com>"
```
to `picky-asn-x509" and
```
"Ionut Mihalcea <ionut.mihalcea@arm.com>"
"Kim Altintop <kim@monadic.xyz>"
```
to `picky`. :slightly_smiling_face:

awakecoding:
@Geobert just to clarify about the copyright - Rust projects normally use the Cargo.toml file to declare authors, so if we add a Contributor.txt file, we'd have to maintain the list of authors in two places. Is there a reason why declaring authors in the Cargo.toml file would not be suitable?

Geobert:
# Copyright

> 
> 
> @Geobert just to clarify about the copyright - Rust projects normally use the Cargo.toml file to declare authors, so if we add a Contributor.txt file, we'd have to maintain the list of authors in two places. Is there a reason why declaring authors in the Cargo.toml file would not be suitable?

Thank you for your answer! I think having the copyright in Cargo.toml is perfect :) I totally forgot this field  ^^'

# ApplicationTag

I tried `pub attributes: ApplicationTag0<Asn1SetOf<Attribute>>,` 

which leads to:
```rust
pub fn new(subject: Name, subject_public_key_info: SubjectPublicKeyInfo) -> Self {
        // It shall be 0 for this version of the standard.
        Self {
            version: 0,
            subject,
            subject_public_key_info,
            attributes: ApplicationTag0(Asn1SetOf(Vec::new())),
        }
    }
```

but this gives:
image

The SET tag is not replaced with the ApplicationTag0 tag :-/

Geobert:
From what Benoît said,  my understanding of ApplicationTag is that its tag replace the tag of the object it contains, so `ApplicationTag0<Asn1SetOf<Attribute>>` should serialize by replacing the 0x31 tag from the SET by 0xA0 from the ApplicationTag but we can see that it is not happening. 

I'm trying to understand how ApplicationTag0 is serialized and I can only see in `ser::Serializer::serialize_newtype_struct` that in the case of Tag0, we "encapsulate" which I'm not sure to understand :-/

Is the ApplicationTag* ser broken? This seems unlikely as you use the crate with no issue and have plenty of tests to back that up. So I'm confused :-/

CBenoit:
> From what Benoît said, my understanding of ApplicationTag is that its tag replace the tag of the object it contains, so `ApplicationTag0<Asn1SetOf<Attribute>>` should serialize by replacing the 0x31 tag from the SET by 0xA0 from the ApplicationTag but we can see that it is not happening.
> 
> I'm trying to understand how ApplicationTag0 is serialized and I can only see in `ser::Serializer::serialize_newtype_struct` that in the case of Tag0, we "encapsulate" which I'm not sure to understand :-/
> 
> Is the ApplicationTag* ser broken? This seems unlikely as you use the crate with no issue and have plenty of tests to back that up. So I'm confused :-/

Oh, let me check, maybe something is wrong somewhere.

Geobert:
> Oh, let me check, maybe something is wrong somewhere.

I don't know if it can help, but I noticed that `is_context_specific` is called in `h_write_header` but I can't find any call to `is_application` anywhere

CBenoit:
> > Oh, let me check, maybe something is wrong somewhere.
> 
> I don't know if it can help, but I noticed that `is_context_specific` is called in `h_write_header` but I can't find any call to `is_application` anywhere

I think I misunderstood something :facepalm: 
Tag is replaced only in marked as IMPLICIT. If EXPLICIT, application tag is basically wrapping similar to a SEQUENCE.
Apparently I implemented application tags as always EXPLICIT and context specific tags as always IMPLICIT.
Sadly, we need an "IMPLICIT" 0xA0. I think I'll need to perform a refactor at some point to fix that issue. For now, let's use something like `Implicit<Option<ApplicationTag0<Attribute>>>`. If it's okay with you I'll push modifications to make it work until I refactor properly.

CBenoit:
> EDIT: by modifying in `ser::Serializer::h_write_header` and adding the testing of `is_application`:
> 
> ```
> Some(last_encapsulator_tag)
>                 if last_encapsulator_tag.is_context_specific() || last_encapsulator_tag.is_application() =>
>             {
>                 written = self.h_write_encapsulator(len)?;
>             }
> ```
> 
> the serialization seems correct, but for a weird reason, it breaks the deserialization:
> 
> ```
> thread 'certification_request::tests::deserialize_csr' panicked at 'failed certification_request_info deserialization: TruncatedData',
> ```

Yes, deserializer would need to be modified the same way. However, as said above, issue is a bit deeper here :confused:

Geobert:
> However, as said above, issue is a bit deeper here

In that case, the PR is complete with `Implicit<Option<ApplicationTag0<Attribute>>>` instead of `Implicit<ApplicationTag0<Asn1SetOf<Attribute>>>`  which needs refactoring to work :)

CBenoit:
> > However, as said above, issue is a bit deeper here
> 
> In that case, the PR is complete with `Implicit<Option<ApplicationTag0<Attribute>>>` instead of `Implicit<ApplicationTag0<Asn1SetOf<Attribute>>>` which needs refactoring to work :)
> 
> The unit test for extensions works, but the empty ApplicationTag0 test is broken of course

Good! I'll fix that, merge and publish :slightly_smiling_face:

CBenoit (approved):
I implemented a workaround so this behaves like a proper `IMPLICIT [0]` on a sequence.

CBenoit (approved):

PR #90, opened 2021-07-02
Maintenance

CBenoit:


ci/azure-pipelines.yaml
```diff
@@ -88,3 +88,22 @@ jobs:
           pathToPublish: $(Build.ArtifactStagingDirectory)
           artifactName: picky
 
+  - job: MSRV
+    displayName: "Minimum supported Rust version check for published crates (if this break, bump crate version minor number)"
```
thenextman, line 92: nit: this is a long name to display in the pipelines UI. You might be better served with a shorter display name and a comment.

thenextman (approved):
LGTM

ci/azure-pipelines.yaml
CBenoit: I agree. I'll do that!

PR #94, opened 2021-08-10
More robust certification validation in `picky` crate

CBenoit:
We noticed a few Authenticode certificates where encoded using a constructed (explicit)
context tag instead of a primitive (implicit) context tag for the subject alternative name
extension (notably PSDiagnostics PowerShell module).

Relevant documentation from RFC5280 Appendix A.2:
```
DEFINITIONS IMPLICIT TAGS ::=

[…]

GeneralName ::= CHOICE {
     otherName                 [0]  AnotherName,
     rfc822Name                [1]  IA5String,
     dNSName                   [2]  IA5String,
     x400Address               [3]  ORAddress,
     directoryName             [4]  Name,
     ediPartyName              [5]  EDIPartyName,
     uniformResourceIdentifier [6]  IA5String,
     iPAddress                 [7]  OCTET STRING,
     registeredID              [8]  OBJECT IDENTIFIER }
```
[Link](https://datatracker.ietf.org/doc/html/rfc5280#appendix-A.2)

`DEFINITIONS IMPLICIT TAGS ::=` is used to specify that except stated otherwise, tags are
implicits (also said primitives).

Picky is encoding this using an implicit context tag as specified by the RFC, and this is a
problem when validating some Windows certificates because picky 6.3.0 (and prior) is fully
parsing the certificate and encode back into der when validating the signature causing
signature validation to fail.

To improve validation robustness, it was decided starting picky 6.4.0 to use the originally
parsed DER representation internally instead of re-encoding on demand.

The aforementioned PSDiagnostics module certificate chain is used as test case to validate
this behavior.


awakecoding (approved):
LGTM

PR #99, opened 2021-09-29
Fix Authenticode bugs that were found by QA

RRRadicalEdward:
Fix that we cannot sign with just an end-entry self-signed root certificate.
Fix that `output` CLI parameter is required for verifying a binary.

### DIFF
--- a/picky/src/x509/pkcs7/authenticode.rs
+++ b/picky/src/x509/pkcs7/authenticode.rs
@@ -179,18 +179,23 @@ impl AuthenticodeSignature {
 
         // certificates contains the signer certificate and any intermediate certificates,
         // but typically does not contain the root certificate
-        let certificates = certificates
-            .into_iter()
-            .filter_map(|cert| {
-                if cert.ty() != CertType::Root {
-                    Some(Certificate::from(cert))
-                } else {
-                    None
-                }
-            })
-            .collect::<Vec<Certificate>>();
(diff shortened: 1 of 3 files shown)


picky/src/x509/pkcs7/authenticode.rs
```diff
@@ -179,18 +179,23 @@ impl AuthenticodeSignature {
...
+                    } else {
+                        None
+                    }
+                })
+                .collect::<Vec<Certificate>>()
+        };
 
-        let signing_cert = certificates.get(0).ok_or(AuthenticodeError::NoCertificates)?;
+        let signing_cert = certificates.get(0);
+        let signing_cert = signing_cert.ok_or(AuthenticodeError::NoCertificates)?;
```
CBenoit, line 198: Not a big deal, but I feel it was better the way it already was (`certificates.get(0).ok_or(AuthenticodeError::NoCertificates)?`).

CBenoit (approved):
Thank you! :smile:

picky/src/x509/pkcs7/authenticode.rs
RRRadicalEdward: Let me change

RRRadicalEdward: Changed

CBenoit:
Thank you!

PR #103, opened 2021-11-02
Implement `time 0.3` types conversion

CBenoit:
I also migrated to 2021 editions `picky`, `picky-server` and `picky-signtool`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,7 +38,7 @@ jobs:
             runner: windows-2019
           - os: linux
             runner: ubuntu-18.04
-          
+
     steps:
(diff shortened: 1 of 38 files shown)


thenextman (approved):
Changes LGTM; CI build is failing on msrv

CBenoit:
> Changes LGTM; CI build is failing on msrv

Looks like we can't check MSRV for crates with different edition versions (checking that picky-asn1 still compiles with 1.43 requires cargo to recognize the 2021 value for the edition field in other crates of workspace…). It's almost like we can't have different editions in our workspace even though it's supposed to work well (and it does until this).
I'll see what we do about that tomorrow.

PR #105, opened 2021-11-09
Add release workflow

thenextman:
Initial release workflow. Details still need to be finalized, but the yml file must exist on main branch to be able to run and iterate on it.


awakecoding (approved):
LGTM

PR #108, opened 2021-11-12
Initial C API and C# .NET wrapper

CBenoit:
This is the C API and C# .NET wrapper MVP implementation.

The Rust files are the single source of truth, everything else is generated and checked against it.

I chose to expose the `picky::pem` module as a first try because it is small and complex enough.

Next modules to expose are:
-  `picky::key`,
- `picky::hash`,
- `picky::jose`,
- `picky::http`,
-  and `picky::x509`.

On C# side, the last remaining bit is to write the NuGet manifest and tell where native libraries are located. Also, I hardcoded path to `./target/debug/libpicky.so` in the `.csproj`, but we'll need to package binaries at least for x64 and x86 architectures (probably arm64 too?) for all supported platforms (Windows and Linux only? Do we support Mac as well?).

`c-api`'s [README](https://github.com/Devolutions/picky-rs/tree/374d1a82c7bfa28e720360172949f51a31d91f7b/c-api) also contains a few additional information.

@thenextman @awakecoding @fdubois1 any thoughts?


c-api/dotnet/src/Picky/Error.cs
```diff
@@ -0,0 +1,39 @@
+namespace Devolutions.Picky;
+
+public class ErrorException : System.Exception
+{
+    public ErrorException(String message) : base(message) { }
```
thenextman, line 5: nit: It's more idiomatic to use the `string` alias than `String`. You can do what you like best, however! But be advised that StyleCop won't like that if you start using it (it only wants you to use `String` when calling static functions on the `String` class...)

c-api/dotnet/src/Picky/Error.cs
```diff
@@ -0,0 +1,39 @@
...
+public static class Error
+{
+    /// <summary>Returns a String containing last error message from Picky.</summary>
+    public static String Last()
+    {
+        int len = Native.Raw.last_error_length_utf8();
+
+        if (len == 0)
+        {
+            throw new ErrorException("no error message to retrieve from Picky");
```
thenextman, line 17: Is this exceptional? It's a public function, I wouldn't `throw` in this case. Returning null has it's own warts but .NET developers love `StringIsNullOrEmpty` to deal with that. Otherwise, you could keep the exception but maybe expose a `HasError` or similar function too, that I can check before calling `Last`.

c-api/dotnet/src/Picky/Native/NativeInheritanceAttribute.cs
```diff
@@ -0,0 +1,23 @@
...
+using System.Diagnostics;
+
+namespace Devolutions.Picky.Native
+{
+    /// <summary>Defines the base type of a struct as it was in the native signature.</summary>
+    [AttributeUsage(AttributeTargets.Struct, AllowMultiple = false, Inherited = true)]
+    [Conditional("DEBUG")]
+    internal sealed partial class NativeInheritanceAttribute : Attribute
+    {
+        private readonly string _name;
```
thenextman, line 11: nit: Actually I prefer the `_` prefix for fields instead of using `this.`. Again, I say do what you like best but StyleCop will probably complain about this if you use it in the future.

c-api/dotnet/src/Picky/Native/Raw.cs
```diff
@@ -0,0 +1,59 @@
+using System.Runtime.InteropServices;
```
thenextman, line 1: Question: this is generated?

c-api/dotnet/src/Picky/Native/VtblIndexAttribute.cs
```diff
@@ -0,0 +1,23 @@
+using System;
```
thenextman, line 1: Is this used anywhere?

c-api/dotnet/src/Picky/Native/NativeInheritanceAttribute.cs
```diff
@@ -0,0 +1,23 @@
+using System;
```
thenextman, line 1: Is this used anywhere?

c-api/dotnet/src/Picky/Native/picky_status.cs
```diff
@@ -0,0 +1,8 @@
+namespace Devolutions.Picky.Native
+{
+    public enum picky_status
```
thenextman, line 3: nit: Is this generated? If so, no big deal. Otherwise, convention would be `PickyStatus` with values `Success` and `Failure`

c-api/dotnet/src/Picky/Pem.cs
```diff
@@ -0,0 +1,170 @@
+﻿namespace Devolutions.Picky;
+
+public class PemException : System.Exception
```
thenextman, line 3: One more time: I think it's fine, but StyleCop would moan about this not being in a separate file

c-api/dotnet/src/Picky/Pem.cs
```diff
@@ -0,0 +1,170 @@
...
+
+            return buf;
+        }
+    }
+
+    ~Pem()
+    {
+        unsafe
+        {
+            Native.Raw.pem_drop(this.inner);
```
thenextman, line 166: I don't know the semantics behind the native object; do you want this to be deterministic? Your native object won't be dropped until the GC runs. If you want it to be deterministic, you can implement `IDisposable` - the end-user can then use `try`/`finally` or `using` or just call `Dispose` to get rid of the native reference *immediately*.

c-api/dotnet/src/Picky/Picky.csproj
```diff
@@ -0,0 +1,20 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+  </PropertyGroup>
```
thenextman, line 8: It's possible to provide the nuget metadata directly in the project, and add `GeneratePackageOnBuild` - you'll get a nuget package automatically on build.

thenextman (approved):
LGTM! I made a few comments mainly around conventions - ultimately I think you should proceed as you see fit, but it's nice to have consistency on things (e.g. `String` vs `string`, leading underscores or not in field names, etc)

You don't necessarily need a separate nuget manifest, it's possible to specify most of the metadata directly in the project

Message me when you're ready and I'll show you the structure for wrapping the native libs across platform, etc

c-api/dotnet/src/Picky/Native/Raw.cs
CBenoit: Yes everything in the `Native` folder is generated

c-api/dotnet/src/Picky/Native/NativeInheritanceAttribute.cs
CBenoit: I can definitely change that (I remember doing something similar in C++ with a `m_` prefix)

c-api/dotnet/src/Picky/Error.cs
CBenoit: That was actually the initial design, but I changed because I don't want the user to check "native" errors that way (a `HasError` function would have the same issue).
The reason is that if user forget, or deliberately _don't_ call the function after calling a native function that could have failed, the thread local static don't get cleared and another native call later that didn't failed could be considered as failed because of the error of the previous call.
As such, user is supposed to check if the native function returned a special value (null, -1…) and only then call this function to retrieve the error description. So I figured that maybe an exception would be appropriate.
What do you think? Maybe I should explain this in the function doc?

c-api/dotnet/src/Picky/Error.cs
CBenoit: Thank you! That's my first code in C# so I'm not really aware of the idioms, I'll definitely follow you on that

c-api/dotnet/src/Picky/Native/VtblIndexAttribute.cs
CBenoit: (generated code)

c-api/dotnet/src/Picky/Native/NativeInheritanceAttribute.cs
CBenoit: (generated code)

c-api/dotnet/src/Picky/Native/picky_status.cs
CBenoit: Yes, this is generated code from C header and it should not be used in idiomatic C# code but only when calling native functions (I'll write new C# enums when appropriate)

c-api/dotnet/src/Picky/Pem.cs
CBenoit: *remembers bad times with Java* I'll just move in a separate file :+1:

c-api/dotnet/src/Picky/Pem.cs
CBenoit: It's just a `picky::pem::Pem` on the heap, as long as it gets dropped at some point it's fine. I figured it was more ergonomic from a C# user point of view to just let the GC do the job here, but what do you think?

c-api/dotnet/src/Picky/Picky.csproj
CBenoit: Oh nice! I'll poke you again about that then

c-api/dotnet/src/Picky/Pem.cs
CBenoit: Oh, maybe I can define the exception class _inside_ the `Pem` class? Would that be idiomatic enough?

c-api/dotnet/src/Picky/Pem.cs
thenextman: @CBenoit Yes it's totally fine as long as it doesn't need to be deterministic. I don't know the internals of `Pem` so I wasn't sure!

c-api/dotnet/src/Picky/Pem.cs
thenextman: Nested public classes tends to be a bit of a code smell, although there's nothing *really* wrong with it.

I'd leave it scoped the same, and move to a separate file or just leave as-is.

c-api/dotnet/src/Picky/Error.cs
thenextman: > I don't expect user to use this often as we provide idiomatic wrappers around the native calls anyway

In that case, I would leave it as is!

c-api/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+[lib]
+name = "picky"
+crate-type = ["staticlib", "cdylib"]
+doc = false
+test = false
+doctest = false
+
+[dependencies]
+picky = { version = "6.4.0", default-features = false, features = ["x509", "jose", "http_signature", "time_conversion"], path = "../picky" }
+ffi_helpers = { version = "0.3", git = "https://github.com/CBenoit/ffi_helpers", branch = "error_format_anyhow_debug" }
```
thenextman, line 16: Any plan to use the published crate, sync this up or make a Devolutions fork? Maybe add a #TODO at least. Just a thought.

c-api/dotnet/.editorconfig
```diff
@@ -0,0 +1,364 @@
+root = true
```
thenextman, line 1: Do you want to commit this?

c-api/dotnet/src/Picky/Picky.csproj
```diff
@@ -0,0 +1,37 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <Company>Devolutions</Company>
+    <Description>Bindings to Rust picky native library</Description>
+    <TargetFramework>net6.0</TargetFramework>
```
thenextman, line 6: Did you check this is good for DVLS? Note that you can target multiple frameworks if needed (`<TargetFrameworks>{comma-seperated list of frameworks}</TargetFrameworks>`)

thenextman (approved):
I added a couple of small comments but LGTM

c-api/dotnet/src/Picky/Picky.csproj
awakecoding: DVLS is still using the .NET Framework so we'd need at least to support the old and the new .NET runtime in the project file

thenextman: @CBenoit See [here](https://docs.microsoft.com/en-us/dotnet/standard/frameworks)

c-api/dotnet/.editorconfig
CBenoit: The `.editorconfig` file? Yes, that's how `dotnet format` is configured (it's also not bad for any editor supporting editorconfig)

c-api/Cargo.toml
CBenoit: Oh, yeah! My patch got merged upstream, but it's not yet published. I'll add a TODO + point directly on the original repo. :+1:

c-api/dotnet/.editorconfig
thenextman: Oh ok!

PR #109, opened 2021-11-26
c-api: expose hash API

CBenoit:
Refine a bit bindgen workflow and expose the `picky::hash` module to C and C# wrappers (the `HashAlgorithm` enum is required by other modules including `ssh`).


thenextman (approved):
LGTM

PR #111, opened 2021-12-01
Fix SSH certificate signature generation

RRRadicalEdward:
I made an error that `cert_key_type` was deserialized twice while computing an `SSH` certificate signature while rebasing the ssh-ca branch to the latest master. This makes our `SSH` certificates' signatures incorrect. This PR fixes this.


CBenoit (approved):
LGTM :+1:

PR #112, opened 2021-12-01
Authenticode timestamping bug fixes

RRRadicalEdward:
1. picky-signtool:
- Fix that `LIEF` logs can appear while verifying a binary(even with binary logging turn off).
- Fix incorrect behavior of `ca` flags
2. picky Authenticode related code:
- Fix a chain verification if the chain also contains timestamp certificates
- Add parsing of `GeneralName::OtherName` as missing it causes incorrect validation of some binaries.

### DIFF
--- a/picky-asn1-x509/src/name.rs
+++ b/picky-asn1-x509/src/name.rs
@@ -1,5 +1,5 @@
 use crate::{AttributeTypeAndValue, AttributeTypeAndValueParameters, DirectoryString};
-use picky_asn1::tag::{Encoding, TagClass, TagPeeker};
+use picky_asn1::tag::{Encoding, Tag, TagClass, TagPeeker};
 use picky_asn1::wrapper::*;
 use serde::{de, ser, Deserialize, Serialize};
 use std::fmt;
@@ -183,7 +183,7 @@ pub type GeneralNames = Asn1SequenceOf<GeneralName>;
 /// ```
 #[derive(Debug, PartialEq, Clone)]
 pub enum GeneralName {
-    //OtherName(OtherName),
+    OtherName(OtherName),
     Rfc822Name(IA5StringAsn1),
     DnsName(IA5StringAsn1),
     //X400Address(ORAddress),
@@ -219,6 +219,11 @@ impl ser::Serialize for GeneralName {
         S: ser::Serializer,
     {
         match &self {
+            GeneralName::OtherName(name) => {
+                let mut raw_der = picky_asn1_der::to_vec(name).map_err(ser::Error::custom)?;
+                raw_der[0] = Tag::context_specific_constructed(0).inner();
+                picky_asn1_der::Asn1RawDer(raw_der).serialize(serializer)
+            }
             GeneralName::Rfc822Name(name) => ImplicitContextTag1(name).serialize(serializer),
             GeneralName::DnsName(name) => ImplicitContextTag2(name).serialize(serializer),
             GeneralName::DirectoryName(name) => ImplicitContextTag4(name).serialize(serializer),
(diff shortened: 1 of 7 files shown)


picky-asn1-x509/src/name.rs
```diff
@@ -341,8 +346,63 @@ impl<'de> de::Deserialize<'de> for GeneralName {
...
+        buff.extend_from_slice(
+            &picky_asn1_der::to_vec(&self.type_id).map_err(|e| ser::Error::custom(format!("{:?}", e)))?,
+        );
+        buff.extend_from_slice(
+            &picky_asn1_der::to_vec(&self.value).map_err(|e| ser::Error::custom(format!("{:?}", e)))?,
+        );
+
+        ExplicitContextTag0::from(Asn1RawDer(buff)).serialize(serializer)
+    }
+}
```
CBenoit, line 371: Can't this be derived?

picky-asn1-x509/src/name.rs
```diff
@@ -341,8 +346,63 @@ impl<'de> de::Deserialize<'de> for GeneralName {
...
+                    "GeneralName::OtherName value"
+                );
+
+                Ok(OtherName { type_id, value })
+            }
+        }
+
+        deserializer.deserialize_seq(Visitor)
+    }
+}
```
CBenoit, line 405: Can't this be derived?

CBenoit (commented):
LGTM! Just a few questions (see comments).

picky-asn1-x509/src/name.rs
RRRadicalEdward: This one can't be. `SEQUENCE` tag is replaced with `ExplicitContextTag0`.
image
Making `Serialize` derived would force the `SEQUENCE` tag to be present. Is there a proper solution?

CBenoit: I understand the `SEQUENCE` tag must indeed be present as per RFC 4043:
image
As such, the following (the one in your message) is appropriate encoding:
image
(not sure if you meant something else?)

CBenoit: Oooh, sorry, I understand! You meant as opposed to this:
image
Technically, that's how it should be encoded standalone:
```
// OtherName ::= SEQUENCE {
//      type-id    OBJECT IDENTIFIER,
//      value      [0] EXPLICIT ANY DEFINED BY type-id}
```

But you're right in that the `GeneralName` choice use an _IMPLICIT_ [0] tag to wrap that. This means that when serializing a `GeneralName` specifically we have to write the [0] tag instead of the `SEQUENCE` tag.

I suggest you just derive `Serialize` on `OtherName`, and modify the manual `Serialize` implementation of `GeneralName` in the "other name" match arm like so:
```rust
        let mut raw_der = picky_asn1_der::to_vec(&self.0).unwrap_or_else(|_| vec![0]);
        raw_der[0] = Tag::context_specific_constructed(0).inner();
        picky_asn1_der::Asn1RawDer(raw_der).serialize(serializer)
```
(still a workaround because we have no proper way to work with that as of now)

Since I did the modification locally to make sure it worked, I just committed it directly: https://github.com/Devolutions/picky-rs/pull/112/commits/bddc0df9cec61717c6205c83aa4f63eff661adf3

RRRadicalEdward: Thank you!

CBenoit (approved):
LGTM!

picky-asn1-x509/src/name.rs
RRRadicalEdward: @CBenoit  
Do you think we could add support for such situations in the future? 
Maybe, you should open an issue to not forget about it. Someone could even fix this :slightly_smiling_face:

CBenoit: Yes, I would like a proper solution to this, but it requires a deep refactoring:
- Updating the serde `Serializer` and `Deserializer` implementations (not `Serialize`, `Deserialize` but the actual serializer code). This is quite involved.
- Adding a ASN.1 value type similar to `serde_json::Value` for loose deserialization (which should replace the `RawDerAsn1` use cases).
- Updating the code using the old approach
- I'm not sure how to fix everything yet and probably need a few experimentation

This will be required when we add BER support (for one, we can't depend on the `RawDerAsn1` in this case), so I plan on doing that on this occasion :slightly_smiling_face: (or sooner if I get the chance, though Devolutions Gateway is taking my time right now)

PR #113, opened 2021-12-01
Fix signing UTF-8/16 powershell scripts

RRRadicalEdward:
Bug fixes _PowerShell_ scripts signing:
- Fix signing _UTF16_ LE/BE _PowerShell_ scripts. 
- If a character in a _PS_ script is not valid, encode it as a _REPLACEMENT CHARACTER_ (�) for file hash calculation. It's how _singtool_ does. Previously signing a script with an invalid character causes the signature to be invalid, although _signtool_ can handle such a script. The bug was discovered with [PSLogging](https://github.com/9to5IT/PSLogging).


picky-signtool/src/utils.rs
```diff
@@ -0,0 +1,4 @@
+pub fn str_to_utf16_bytes(s: &str, buff: &mut Vec<u8>) {
+    s.encode_utf16()
+        .for_each(|word| buff.extend_from_slice(&word.to_le_bytes()));
+}
```
CBenoit, line 4: This could be written as:
```suggestion
pub fn str_to_utf16_bytes(s: &str, buff: &mut Vec<u8>) {
    buff.extend(s.encode_utf16().map(|word| word.to_le_bytes()).flatten())
}
```

picky-signtool/src/verify.rs
```diff
@@ -154,15 +158,16 @@ pub fn authenticode_signature_ps_from_file(file_path: &Path) -> anyhow::Result<A
...
+    let (buffer, ..) = if buffer.starts_with(&UTF16_LE_BOM) {
+        encoding_rs::UTF_16LE.decode(&buffer)
+    } else if buffer.starts_with(&UTF16_BE_BOM) {
+        encoding_rs::UTF_16BE.decode(&buffer)
+    } else {
+        encoding_rs::UTF_8.decode(&buffer)
+    };
 
-    let signature = extract_ps_authenticode_signature(buffer)
+    let signature = extract_ps_authenticode_signature(buffer.to_string())
```
CBenoit, line 170: If `extract_ps_authenticode_signature` absolutely requires an owned String, then use `into_owned` so that there is no copy if the value is already owned. But looking at the `extract_ps_authenticode_signature` function, I see this could take a `&str` instead.

picky-signtool/src/verify.rs
```diff
@@ -244,14 +249,29 @@ fn apply_flags<'a>(
     }
 }
 
-pub fn extract_signed_ps_file_content(raw_content: String) -> String {
-    let end = match raw_content.find(PS_AUTHENTICODE_HEADER) {
-        Some(index) => index - 2, // -2 to remove \r\n from the `# SIG # Begin signature block` line
+pub fn extract_signed_ps_file_content_bytes(raw_content: &[u8]) -> &[u8] {
+    // raw_content - utf-16 encoded bytes
+    let find = |haystack: &[u8], needle: &Vec<u8>| -> Option<usize> {
```
CBenoit, line 254: Taking `&[u8]` is always better than `&Vec<u8>` when possible
```suggestion
    let find = |haystack: &[u8], needle: &[u8]| -> Option<usize> {
```

CBenoit (changes requested):
I love when `Cargo.lock` get slimmer :laughing: 
So this is looking pretty good! I just have a few minor suggestions and this can be merged!

CBenoit (approved):
LGTM! Thank you 😄

PR #114, opened 2021-12-17
Add Access-Control-Max-Age to cache CORS preflight request responses

sduquette-devolutions:
When making a cross-origin request, browsers need to make a preflight request (with the OPTIONS verb) to check if a CORS header is allowing it. The Access-Control-Max-Age lets browser cache the preflight response, reducing the number of OPTIONS requests that are needed. The value is set to 2 hours (7200 seconds) which is the maximum value allowed by Chrome.


CBenoit (approved):
LGTM, thank you!

PR #116, opened 2022-01-28
DPS-5446: expose SSH certificates API to C#

CBenoit:
New approach for bindings based on [these experiments](https://github.com/CBenoit/dotnet-to-rust-bindgen-experiments).

cc @awakecoding


thenextman (approved):
LGTM

Issue #91, opened 2021-07-24
ApplicationTagN bits are incorrect


Albibek:
Hi.

I think I've found a bug in ASN.1 implementation for the application tag. 
The standard requires that tag class must be specified in bits 7 and 8 while bit 6 is reserved for constructed/primitive definition.

Here are the links at the simplified explanations:
http://luca.ntop.org/Teaching/Appunti/asn1.html
https://letsencrypt.org/docs/a-warm-welcome-to-asn1-and-der/

The actual code defines application tag bits as [0xA0](https://github.com/Devolutions/picky-rs/blob/704dc1439dfcc5209fe2eda498ea6dedd06532b9/picky-asn1/src/tag.rs#L61) which is 0b10100000 binary and represents a context-specific tag with the constructed bit set. I believe it should be 0x40 (0b01000000) instead. 

It definitely affects tag class matching because of overlapping intervals also making it match order dependent, and this is how I found it. It is also most probably the reason of using multiple tag class confusions for tagged structures like [here](https://github.com/Devolutions/picky-rs/blob/704dc1439dfcc5209fe2eda498ea6dedd06532b9/picky-asn1-x509/src/name.rs#L265), because the whole x509 seems to be using only context specific tags.

In addition, I see some small edge case problem for code doing an OR operation in `Tag::application` and `Tag::context_specific`. For example, passing 254 into one of these constructors will not make a tag class to be correctly set to Application/Context specific leaving it Private instead. To avoid such situation, the higher bits should be cleared first (using 0x3F = 0b00111111) to guarantee the class, or, some numbers should probably not be accepted at all.
```
    #[inline]
    pub const fn application(number: u8) -> Self {
        Tag(number & 0x3f | 0x40)
    }
    
    #[inline]
    pub const fn context_specific(number: u8) -> Self {
        Tag(number & 0x3f | 0x80)
    }

```

Along with fixing I want to propose changing `Tag::is_application` and `Tag::is_context_specific` methods to bit-based comparisons as well, so they are more semantically accurate and possibly a small bit more performant due to less branching.
```
     #[inline]
    pub const fn is_application(self) -> bool {
      self.0 & 0xC0 == 0x40
    }

    #[inline]
    pub const fn is_context_specific(self) -> bool {
         self.0 & 0xC0 == 0x80
    }
```

CBenoit:
Hi @Albibek, thank you for opening the issue!
[...]

PR #117, opened 2022-01-31
KDC proxy in DG related improvements

RRRadicalEdward:
KDC proxy in DG related improvements:
- Added `ApplicationTag<V, const T: u8>` for easier using needed `ApplicationTagN` without adding a lot of new `ApplicationTag`s.
- Added `GENERAL_STRING`.
- Added `impl<T: Default> Default for Optional<T>`.
- Added clearing `encapsulator_tag_stack` as otherwise, it would contain unwanted tags on the next serialization.

### DIFF
--- a/picky-asn1-der/src/application_tag.rs
+++ b/picky-asn1-der/src/application_tag.rs
@@ -1,0 +1,158 @@
+use crate::misc::Length;
+use crate::Asn1RawDer;
+use picky_asn1::tag::{Tag, TagPeeker};
+use serde::de::{Error, SeqAccess};
+use serde::{de, ser};
+use std::fmt;
+use std::fmt::Debug;
+
+#[derive(Debug, PartialEq)]
+pub struct ApplicationTag<V: Debug + PartialEq, const T: u8>(V);
+
+impl<V: Debug + PartialEq, const T: u8> ApplicationTag<V, T> {
+    pub fn from(value: V) -> Self {
+        Self(value)
+    }
+}
+
+#[derive(Debug)]
+struct ApplicationTagInner<V: Debug>(V);
+
+impl<'de, V: de::Deserialize<'de> + Debug + PartialEq> de::Deserialize<'de> for ApplicationTagInner<V> {
+    fn deserialize<D>(deserializer: D) -> Result<Self, <D as de::Deserializer<'de>>::Error>
+    where
+        D: de::Deserializer<'de>,
+    {
+        struct Visitor<E>(Option<E>);
+
+        impl<E> Visitor<E> {
+            pub fn new() -> Self {
+                Self(None)
+            }
+        }
+
+        impl<'de, E: de::Deserialize<'de> + Debug> de::Visitor<'de> for Visitor<E> {
+            type Value = ApplicationTagInner<E>;
+
+            fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+                formatter.write_str("Error inner")
+            }
+
+            fn visit_seq<A>(self, mut seq: A) -> Result<Self::Value, A::Error>
+            where
+                A: SeqAccess<'de>,
+            {
+                let res: E = seq
+                    .next_element()?
+                    .ok_or(A::Error::missing_field("ApplicationTagInner"))?;
+
+                Ok(ApplicationTagInner(res))
+            }
+        }
+
+        deserializer.deserialize_seq(Visitor::<V>::new())
+    }
+}
+
+impl<'de, V: de::Deserialize<'de> + Debug + PartialEq, const T: u8> de::Deserialize<'de> for ApplicationTag<V, T> {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: de::Deserializer<'de>,
+    {
+        struct Visitor<E>(Option<E>, u8);
+
+        impl<E> Visitor<E> {
+            pub fn new(tag: u8) -> Self {
+                Self(None, tag)
+            }
+        }
+
+        impl<'de, E: de::Deserialize<'de> + Debug + PartialEq> de::Visitor<'de> for Visitor<E> {
+            type Value = E;
+
+            fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+                formatter.write_str("Error inner")
+            }
+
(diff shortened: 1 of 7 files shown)


picky-asn1/src/wrapper.rs
```diff
@@ -200,6 +200,7 @@ asn1_wrapper! { auto struct IA5StringAsn1(IA5String),               Tag::IA5_STR
 asn1_wrapper! { auto struct BMPStringAsn1(BMPString),               Tag::BMP_STRING }
 asn1_wrapper! { auto struct UTCTimeAsn1(UTCTime),                   Tag::UTC_TIME }
 asn1_wrapper! { auto struct GeneralizedTimeAsn1(GeneralizedTime),   Tag::GENERALIZED_TIME }
+asn1_wrapper! { auto struct GeneralStringAsn1(IA5String),           Tag::GENERAL_STRING }
```
CBenoit, line 203: I see you just wrapped the IA5String which is surprising to me. Is that how the GeneralString is defined?

picky-asn1-der/src/application_tag.rs
```diff
@@ -0,0 +1,158 @@
+use crate::misc::Length;
+use crate::Asn1RawDer;
+use picky_asn1::tag::{Tag, TagPeeker};
+use serde::de::{Error, SeqAccess};
+use serde::{de, ser};
+use std::fmt;
+use std::fmt::Debug;
+
+#[derive(Debug, PartialEq)]
+pub struct ApplicationTag<V: Debug + PartialEq, const T: u8>(V);
```
CBenoit, line 10: I think we already bumped the minimal required rustc version for the next release, but don't forget to take this into account when using latest features :slightly_smiling_face:

picky-asn1-der/src/application_tag.rs
```diff
@@ -0,0 +1,129 @@
...
+            pub fn new(tag: u8) -> Self {
+                Self(None, tag)
+            }
+        }
+
+        impl<'de, E: de::Deserialize<'de> + Debug + PartialEq> de::Visitor<'de> for Visitor<E> {
+            type Value = E;
+
+            fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+                formatter.write_str(&format!("A valid DER-encoded ApplicationTag{}", self.1))
```
CBenoit, line 35:
```suggestion
                formatter.write_str(&format!("A valid DER-encoded ApplicationTag{:02}", self.1))
```
To print the leading zero if applicable

picky-asn1-der/src/application_tag.rs
```diff
@@ -0,0 +1,129 @@
+use crate::misc::Length;
+use crate::Asn1RawDer;
+use picky_asn1::tag::{Tag, TagPeeker};
+use serde::de::{Error, SeqAccess};
+use serde::{de, ser};
+use std::fmt;
+use std::fmt::Debug;
+
+#[derive(Debug, PartialEq)]
+pub struct ApplicationTag<V: Debug + PartialEq, const T: u8>(V);
```
CBenoit, line 10: Also, do you remember https://github.com/Devolutions/picky-rs/pull/92 and https://github.com/Devolutions/picky-rs/issues/91?
If I understand correctly you did a generic implementation of `ExplicitContextTag` (you need to rename from `ApplicationTag`)

picky-asn1-der/src/application_tag.rs
```diff
@@ -0,0 +1,129 @@
...
+        Self(value)
+    }
+}
+
+impl<'de, V: de::Deserialize<'de> + Debug + PartialEq, const T: u8> de::Deserialize<'de> for ApplicationTag<V, T> {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: de::Deserializer<'de>,
+    {
+        struct Visitor<E>(Option<E>, u8);
```
CBenoit, line 23: Why `Option<E>`? Also the tag could stay as a const generic

picky-asn1-der/src/application_tag.rs
```diff
@@ -0,0 +1,129 @@
...
+                .map_err(|e| S::Error::custom(format!("Cannot serialize Application tag inner value: {:?}", e)))?;
+        }
+
+        let mut res = vec![Tag::application_constructed(T).inner()];
+
+        Length::serialize(buff.len(), &mut res)
+            .map_err(|e| S::Error::custom(format!("Cannot serialize Length: {:?}", e)))?;
+        res.extend_from_slice(&buff);
+
+        Asn1RawDer(res).serialize(serializer)
```
CBenoit, line 106: You can use a single buffer

```suggestion
        let mut buf = vec![Tag::application_constructed(T).inner()];

        {
            let mut s = crate::Serializer::new_to_byte_buf(&mut buf);
            self.0
                .serialize(&mut s)
                .map_err(|e| S::Error::custom(format!("Cannot serialize Application tag inner value: {:?}", e)))?;
        }

        Length::serialize(buf.len() - 1, &mut buf)
            .map_err(|e| S::Error::custom(format!("Cannot serialize Length: {:?}", e)))?;
        
        Asn1RawDer(buf).serialize(serializer)
```

picky-asn1-der/src/application_tag.rs
RRRadicalEdward: This is a workaround. It allows me to specify `type Value = E;`(see below) otherwise I should specify `type Value = V` that gives the `error: generic parameter from outer function` compile error. Maybe, you have a better solution. :slightly_smiling_face:

picky-asn1/src/wrapper.rs
RRRadicalEdward: I just took the `GeneralString` definition from the `Kerberos` ASN.1 grammar specification(It's said that `GeneralString` contains `IA5String`).

CBenoit: OK, I guess we support a _subset_ then. Can you add some comment about that?

picky-asn1-der/src/application_tag.rs
RRRadicalEdward: `PhantomData` is an idiomatic workaround 😄

CBenoit (approved):
Looks like all concerns are addressed, thank you :tada:

PR #123, opened 2022-02-07
Use netstandard2.0 for nuget package

awakecoding:


c-api/dotnet/Devolutions.Picky/Devolutions.Picky.csproj
```diff
@@ -3,14 +3,14 @@
   <PropertyGroup>
     <Company>Devolutions</Company>
     <Description>Bindings to Rust picky native library</Description>
-    <TargetFrameworks>netstandard2.0;net472;net48;net6.0</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0</TargetFrameworks>
```
CBenoit, line 6: Supporting multiple targets is an issue?

awakecoding: I don't know why but right now building the package makes it build for one target framework once + copy it into each target-specific directories, rather than rebuild it for each target framework properly. I don't know yet how to fix multi-framework targetting, but since we support netstandard2.0, let's just work around the problem for now.

PR #129, opened 2022-02-15
Fix verifying an Authenticode signature certificates chain with timestamp certificates chain

RRRadicalEdward:
`Authenticode` timestamp protocol that we implemented places the timestamp certificates chain right after the signature certificates chain. But some signature certificates validations didn't handle such situations(or couldn't). So it added a specified `SignatureCertificatesIterator` iterator that filters out timestamp certificates and uses it where we expect only signature certificates. Also, I rewrite some functions to accept an iterator as a parameter.

### DIFF
--- a/picky/src/x509/pkcs7/authenticode.rs
+++ b/picky/src/x509/pkcs7/authenticode.rs
@@ -37,6 +37,7 @@ use picky_asn1_x509::pkcs7::signer_info::{
 use picky_asn1_x509::pkcs7::Pkcs7Certificate;
 use picky_asn1_x509::{oids, AttributeValues, Certificate, DigestInfo, Name};
 use std::cell::RefCell;
+use std::iter::Iterator;
 use std::ops::DerefMut;
 use thiserror::Error;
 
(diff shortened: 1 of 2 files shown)


picky/src/x509/pkcs7/authenticode.rs
```diff
@@ -1159,21 +1191,22 @@ impl<'a> AuthenticodeSignatureBuilder<'a> {
...
+        let signing_certificate = certificates
+            .iter()
+            .find(|cert| {
+                Name::from(cert.issuer_name()) == issuer_and_serial_number.issuer
+                    && cert.serial_number() == &issuer_and_serial_number.serial_number.0
+            })
+            .ok_or(AuthenticodeError::NoCertificatesAssociatedWithIssuerAndSerialNumber {
+                issuer: issuer_and_serial_number.issuer.clone(),
+                serial_number: issuer_and_serial_number.serial_number.0 .0.clone(),
+            })
```
CBenoit, line 1203: This should probably be `ok_or_else` because of the `clone`s (which are deep copies in this case). You may try to run `cargo clippy --all-features` to make sure clippy is checking this module.

(I'm think of [this lint](https://rust-lang.github.io/rust-clippy/master/#or_fun_call))

picky/src/x509/pkcs7/authenticode.rs
```diff
@@ -991,15 +990,15 @@ impl<'a> AuthenticodeValidator<'a> {
...
+
+    if certificates_iter
         .flat_map(|cert| cert.extensions().iter())
         .any(|extension| matches!(extension.extn_value(), ExtensionView::ExtendedKeyUsage(_)))
     {
-        let signing_cert = certificates.first().ok_or(AuthenticodeError::NoCertificates)?;
-        if !signing_cert
-            .extensions()
+        let signing_certificate_extensions = signing_certificate.extensions();
+        if !signing_certificate_extensions
```
CBenoit, line 1001: Not a big deal, but there is probably little value in declaring an intermediate variable in this case.

CBenoit (commented):
Thank you! Small details, but looks good otherwise :+1:

CBenoit (approved):
Thank you! :tada:

PR #132, opened 2022-03-01
Handle verification of keys using elliptic curve

flavio:
This PR introduces support for handling certificates using elliptic curve public keys.

~~Only signature verification is currently implemented.~~
**Update:** Both signature creation and verification are supported.

The implementation relies on the `ring` crate, which currently is the only one capable of supporting both ECDSA p256 and ECDSA p384.

~~I haven't yet extended the unit tests. I'm looking forward to hear your feedback about this contribution. I think this is something you're interested about (see #100). I wonder if this is going in the direction you like.~~
**Update:** The unit tests have been extended too

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1993,6 +1993,7 @@ dependencies = [
  "reqwest 0.11.7",
  "ring",
  "rsa",
+ "rstest",
  "serde",
  "serde_json",
  "sha-1",
@@ -2631,6 +2632,19 @@ dependencies = [
  "rand 0.8.4",
  "subtle",
  "zeroize",
+]
+
+[[package]]
+name = "rstest"
+version = "0.12.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d912f35156a3f99a66ee3e11ac2e0b3f34ac85a07e05263d05a7e2c8810d616f"
+dependencies = [
+ "cfg-if 1.0.0",
+ "proc-macro2",
+ "quote",
+ "rustc_version",
+ "syn",
 ]
 
 [[package]]
--- a/picky/Cargo.toml
+++ b/picky/Cargo.toml
@@ -52,12 +52,14 @@ aes-gcm = { version = "0.9.4", optional = true }
 bcrypt-pbkdf = { version = "0.6", optional = true }
 block-modes = { version = "0.8", optional = true }
 aes = { version = "0.7.5", features = ["ctr"], default-features = false, optional = true }
+ring = "0.16.20"
 
 [dev-dependencies]
 pretty_assertions = "1.0.0"
 hex = "0.4.3"
 cfg-if = "1.0.0"
 ring = "0.16.20"
+rstest = "0.12.0"
 
 [features]
 default = ["x509", "jose", "http_signature", "http_trait_impl"]
--- a/picky/src/http/http_signature.rs
+++ b/picky/src/http/http_signature.rs
@@ -829,6 +829,9 @@ const HTTP_SIG_ALGO_RSA_SHA2_512: &str = "rsa-sha2-512";
 const HTTP_SIG_ALGO_RSA_SHA3_384: &str = "rsa-sha3-384";
 const HTTP_SIG_ALGO_RSA_SHA3_512: &str = "rsa-sha3-512";
 
+const HTTP_SIG_ALGO_ECDSA_SHA_256: &str = "ecdsa-sha256";
+const HTTP_SIG_ALGO_ECDSA_SHA_384: &str = "ecdsa-sha384";
+
 fn to_http_sig_algo_str(algo: SignatureAlgorithm) -> &'static str {
     match algo {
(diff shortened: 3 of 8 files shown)


CBenoit:
Thank you very much :tada: 

> I haven't yet extended the unit tests. I'm looking forward to hear your feedback about this contribution. I think this is something you're interested about (see #100). I wonder if this is going in the direction you like.

This is going in the right direction! However, I wanted to use [`RustCrypto`'s crates](https://github.com/RustCrypto/signatures/tree/master/ecdsa) to stay purely Rust only.
I think both ECDSA p256 and ECDSA p384 are supported:
- https://crates.io/crates/p256/
- https://crates.io/crates/p384/

Would that be acceptable on you side?

I read the issue on `sigstore-rs`. Glad to see `picky` have been considered as a possible candidate for what your trying to do despite the limited popularity. I understand if you wish to switch to [webpki](https://github.com/briansmith/webpki) in the future. In the meantime, I'll be happy to help and merge PR for functionalities you might need. :slightly_smiling_face:

flavio:
> This is going in the right direction! However, I wanted to use [`RustCrypto`'s crates](https://github.com/RustCrypto/elliptic-curves) to stay purely Rust only. I think both ECDSA p256 and ECDSA p384 are supported:

I kinda imagined that :) I started to use `ecdsa` + `p256` + `p384` crates but then I realized the `p384` crate isn't mature enough. It cannot be used yet, because some traits required by the `ecdsa` structs are not implemented yet. This is tracked inside of this issue: https://github.com/RustCrypto/elliptic-curves/issues/240

If you want, I have another branch that builds on top of [ring-compat](https://github.com/RustCrypto/ring-compat). This however would bring `ring` as a dependency. Plus, the code using `ring-compat` looks a bit uglier compared to the one inside of this PR. That's because it's harder to abstract it.

I know `ring` is not written in pure Rust, but at least is not relying on openssl or some other library :)

Right now I need p384 support, because this what Fulcio (a project part of sigstore) uses.

If you are really against using `ring`, I could build EC P256 support on top of `ecdsa` + `p256`. While EC P384 could be hidden behind a feature flag, and be implemented via either `ring` or `ring-compat`. To be honest, this would be a bit painful and the final code would probably look uglier, but this could be a temporary solution until `p384` matures.

What do you think?

thenextman:
Hello @flavio; I resolved an issue in the CI that prevented checks from running on your branch. If you rebase on master the build should now run (tests are failing but I see from your OP that you haven't touched that side yet). Thank you for your contribution.

flavio:
> Thank you for the insight! I think it's good to use ring at least until p384 crate has arithmetic implemented. I agree, it's not worth to use ring-compat in our case either, we don't even rely that much on their traits in the first place

Thanks, I'll go ahead with this PR and look into the unit tests

flavio:
@thenextman, @CBenoit: I've forcefully pushed. The code now handles both signature creation and verification using ECDSA keys. The unit tests have been extended too, I hope you won't mind the new `rstest` dependency, I like how it can be used to reduce the boilerplate.

All the unit tests are green on my machine, clippy is happy too.

Some details:
  * The whole implementation relies on `ring`, as discussed above
  * Signature creation:
    * ECDSA P-256 keys can be used only with sha2 256 hashing algorithm.
    * ECDSA P-384 keys can be used only with sha2 384 hashing algorithm.
    * ECDSA P-512 keys cannot be used to produce signatures. This is a limitation of `ring` and all the other rust crates.
 * Signature verification:
    * A certificate with a ECDSA P-256 public key assumes the sha2 256 algorithm has been used to produce the signature. This is what the official RFCs recommend (I don't remember the number of the RFC unfortunately, I got lost into the rabbit hole).
    * A certificate with a ECDSA P-384 public key assumes the sha2 384 algorithm has been used to produce the signature. This is what the official RFCs recommend (I don't remember the number of the RFC unfortunately, I got lost into the rabbit hole).
    * Signatures produced with a ECDSA P-512 keys cannot be verified. This is a limitation of `ring` and all the other rust crates.

Thanks again for your prompt and welcoming responses!

CBenoit (approved):
Very high quality pull request, thank you!
Introduction of `rstest` is much welcomed. I just didn't know about it yet. :)

CI shows that some code is still failing to compile behind a feature gate. It's not a big deal so I sent a patch directly since you allowed maintainer modifications.

I'll merge and release next version tomorrow :tada:

By the way, next version will be a candidate release, just to see if there is additional breaking changes before committing to picky 7.x. Would it cause any issue on your side?

flavio:
> CI shows that some code is still failing to compile behind a feature gate. It's not a big deal so I sent a patch directly since you allowed maintainer modifications.

Sorry, I didn't notice that while testing locally. Thanks for having fixed it!

> I'll merge and release next version tomorrow :tada:
> 
> By the way, next version will be a candidate release, just to see if there is additional breaking changes before committing to picky 7.x. Would it cause any issue on your side?

That's fantastic news :clap: 
I have no worries about possible API breaking changes, go ahead with them if you need to.

PR #135, opened 2022-03-09
Separate JWT validation from decoding in JOSE API

CBenoit:
This makes API more convenient to first process header and then decide what kind of validation should be applied, or what claims type to deserialize into.


thenextman (approved):
LGTM

PR #143, opened 2022-05-06
Fetch curve info from private_key_algorithm

dpogorzelski:
This fixes the issue described here: https://github.com/Devolutions/picky-rs/issues/100#issuecomment-1118363996

### DIFF
--- a/picky/src/key.rs
+++ b/picky/src/key.rs
@@ -176,26 +176,6 @@ impl<'a> TryFrom<&'a PrivateKey> for EcdsaKeypair<'a> {
                 let private_key_data = private_key.private_key.to_vec();
                 let public_key_data = private_key.public_key.payload_view();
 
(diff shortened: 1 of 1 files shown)


CBenoit:
Thank you! I’ll review this on Monday :slightly_smiling_face:

dpogorzelski:
perhaps checking key type first is better ( `private_key_info::PrivateKeyValue::RSA(_)` )so that a meaningful error is returned when the key is RSA

CBenoit:
> perhaps checking key type first is better ( `private_key_info::PrivateKeyValue::RSA(_)` )so that a meaningful error is returned when the key is RSA

Yeah, I think it would be better with the check. Looks good to me otherwise!

CBenoit (approved):
LGTM! Thank you!

PR #150, opened 2022-05-26
picky-krb: add missing Kerberos name type constants

awakecoding:
Add missing Kerberos name type constants (https://www.rfc-editor.org/rfc/rfc4120#section-7.5.8) which we'll need in sspi-rs to handle NT_PRINCIPAL (1) and NT_ENTERPRISE (10) name types [based on the username formats (netbios versus UPN)](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-kile/6435d3fb-8cf6-4df5-a156-1277690ed59c). We can temporarily just hardcode the value in sspi-rs until the next release of picky-krb.

### DIFF
--- a/picky-krb/src/constants.rs
+++ b/picky-krb/src/constants.rs
@@ -10,8 +10,15 @@ pub mod types {
 
     pub const KRB_ERROR_MSG_TYPE: u8 = 0x1e;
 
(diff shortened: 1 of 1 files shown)


CBenoit (approved):
LGTM

PR #153, opened 2022-06-23
JWE code review

pdugre:
Quick notes:

For the NonceSize and TagSize, the values are hardcoded because the cbc crate is not imported at all, so I cannot fetch it from there.


CBenoit (approved):
Much better! Thank you!

PR #154, opened 2022-06-24
Zeroize private keys

shurizzle:
I implemented `Zeroize` from `zeroize` crate for `IntegerAsn1` so I can zeroize private values on `RsaPrivateKey` and `ECPrivateKey`. The behavior is driven by `zeroize` flag so you can choose.

Thank you in advance.


picky-asn1-x509/src/private_key_info.rs
```diff
@@ -329,6 +339,16 @@ impl RsaPrivateKey {
...
+    #[cfg(feature = "zeroize")]
+    #[deprecated(note = "field is now public")]
+    pub fn into_public_components(mut self) -> (IntegerAsn1, IntegerAsn1) {
+        (
+            std::mem::take(&mut self.modulus),
+            std::mem::take(&mut self.public_exponent),
+        )
+    }
+
+    #[cfg(not(feature = "zeroize"))]
```
CBenoit, line 351: I think it’s best to just make this the default implementation even if `zeroize` feature is not enabled:
- Less code
- The empty "default" `Vec` is optimized and doesn’t trigger memory allocation (when created through `Vec::new` or `Vec::default` as seen [here](https://github.com/rust-lang/rust/blob/481db40311cdd241ae4d33f34f2f75732e44d8e8/library/alloc/src/vec/mod.rs#L424) and [here](https://github.com/rust-lang/rust/blob/f1f721e64014863f41c1a386b04af04c2de25321/library/alloc/src/raw_vec.rs#L64))
- I’m pretty sure this produce the same code in release profile anyway

picky-asn1/src/wrapper.rs
```diff
@@ -387,6 +387,16 @@ impl IntegerAsn1 {
     }
 }
 
+#[cfg(feature = "zeroize")]
+use zeroize::Zeroize;
```
CBenoit, line 391: I would prefer either putting the `use` statement at the top, or do `impl zeroize::Zeroize …` bellow

CBenoit (changes requested):
Thank you for this! Just a few comments to address and it’s ready to be merged 🙂

picky/Cargo.toml
```diff
@@ -86,3 +86,4 @@ ssh = ["byteorder", "aes", "ctr", "cbc", "bcrypt-pbkdf", "lexical-sort"]
 http_trait_impl = ["http"]
 chrono_conversion = ["chrono", "picky-asn1/chrono_conversion"]
 time_conversion = ["time", "picky-asn1/time_conversion"]
+zeroize = ["picky-asn1-x509/zeroize", "picky-asn1/zeroize"]
```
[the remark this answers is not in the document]
shurizzle: I introduced it as a feature because of the behavior, not for the deps. I'll use picky exclusively with this features enabled so it's not a big deal for me to put it as a mandatory behavior. Tell me if you want zeroize in the default features o just use zeroize as mandatory.

picky-asn1-x509/src/private_key_info.rs
shurizzle: Same in picky/src/key.rs.

picky-asn1/src/wrapper.rs
shurizzle: You are right, my bad.

picky/Cargo.toml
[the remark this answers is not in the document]
CBenoit: Yes, I think it's good to have it mandatory

CBenoit (approved):
LGTM, thank you!

PR #162, opened 2022-07-27
fix(wasm): use `bundler` target

CBenoit:
This gives the best user experience when using frameworks such as Angular and React


thenextman (approved):
LGTM

PR #163, opened 2022-07-28
picky-krb: Add constants related to SECBUFFER_CHANNEL_BINDINGS

RRRadicalEdward:


awakecoding (changes requested):
For the 3.2.5.8 AP Exchange reference link, you can use https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-kile/b15648e2-439a-4d04-b8a2-2f34c45690f9 instead of the PDF link from the Microsoft CDN

CBenoit (approved):
LGTM

awakecoding (approved):

PR #168, opened 2022-08-10
build: workspace clean up

CBenoit:
1/ `picky-server` is removed from the workspace for now

2/ `picky-signtool` dependencies are fixed:

- Use a pinned `picky` dependency from crates.io.
- Use a pinned `lief-cwal` dependency from crates.io

Issue: ARC-88

Also, CI is updated.


Sylfwood (approved):

awakecoding (approved):
LGTM

PR #172, opened 2022-08-26
Deploy only platform specific dependencies via nuget

thenextman:
### Fixes for native libraries in nuget package

On macOS, it's better to deploy a single universal binary as a `NativeReference`. This properly sets up the linker, etc, and allows us to consume the library without manually locating it and loading it first. Xamarin.Mac doesn't seem to understand the layout of the `runtimes` directory (and indeed - there is no `osx-universal` .NET RID).

Additionally, the package deploys **all** native libraries for **all** platforms. This isn't an issue on Windows where RDM essentially whitelists files to be installed; however at least macOS is not selective - if a content file is included, it will be packaged in the final .app. I'm not sure about Linux but certainly this can swell the size of the final application unnecessarily. 

I used a static MSBuild property `IsOSPlatform` to only deploy the proper libraries; this works on Windows and macOS and should work on Linux. In the future, additional thought will need to be given for handling additional targets (`IsOSPlatform` does not have a value for Android or iOS). Possibly we could check the `runtimeid` and fallback on `IsOSPlatform` if needed.


awakecoding (approved):
nice work!

CBenoit (approved):
Confirmed working for Linux as well
image

awakecoding (approved):

dion-gionet (approved):

PR #173, opened 2022-08-31
Kerberos crypto algorithms

RRRadicalEdward:
Hi, I added a new crypto module inside the `picky-krb` which contains _Kerberos_ crypto algorithms, and added more key usages in the `constants.rs` module.

List of the implemented algorithms:
- _AES_ cipher group:
   - _AES128-CTS-HMAC-SHA1-96_
   - _AES256-CTS-HMAC-SHA1-96_
   - _AES_ key derivation
- _DES_ cipher group:   
   - _DES3-CBC-SHA1-KD_
   - _DES_ key derivation

All encryption, decryption, and key derivation algorithms are covered with unit tests. Also, I left some small comments to help code understanding.

I used the following library as a reference: 
 - [A Go implementation of Kerberos crypto](https://github.com/jcmturner/gokrb5/tree/master/crypto).
 
 Used Docs:
- [Encryption and Checksum Specifications for Kerberos 5](https://datatracker.ietf.org/doc/html/rfc3961).
- [The Kerberos Network Authentication Service (V5)](https://www.rfc-editor.org/rfc/rfc4120).
- [Advanced Encryption Standard (AES) Encryption for Kerberos 5](https://www.ietf.org/rfc/rfc3962.txt).
- [CTS](https://en.wikipedia.org/wiki/Ciphertext_stealing).


picky-krb/src/crypto/nfold.rs
```diff
@@ -0,0 +1,166 @@
+#[inline]
```
CBenoit, line 1: You don’t need the #[inline] attribute when the function is used inside the same crate (this is only useful for inlining across crates). There is `#[inline(always)]` to override the inlining heuristic, but I advise against.

CBenoit (approved):
Looks good to me overall 🙂 

Maybe you can address the comment but it’s not a real issue

picky-krb/src/crypto/nfold.rs
RRRadicalEdward: Done

CBenoit:
Also, could you bump picky-krb to 0.4.0? Thank you!

PR #180, opened 2022-09-29
Packaged iOS .framework

thenextman:
The iOS universal library is packaged into a .framework by the build, and installed in the final target as a `NativeReference`.

It's necessary to reference the framework itself in `DllImport` statements, the generated bindings are updated accordingly. I added a new target framework for Xamarin.iOS so we have the `__IOS__` definition on that platform. This also means compiling the managed code with `msbuild`: Xamarin.iOS requires targets from the full MSBuild. A Windows machine is also needed in order to use MSBuild 17 for C# 10 support (the generated bindings use C# 10 features like file-scoped namespace declarations).


richerlariviere (approved):

awakecoding (approved):
I approve of the amount of PowerShell used in this :)

CBenoit (changes requested):
Something’s not right with the generated code, did you try to run the dotnet tests?

Info.plist
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
CBenoit, line 1: Do we need this file to be at the root?

thenextman: No, I can move it up to the ffi/dotnet

CBenoit: If it’s dotnet specific, I would prefer. Thank you!

thenextman:
> Something’s not right with the generated code, did you try to run the dotnet tests?

Bindings are regenerated as per our discussion 🤞

CBenoit (approved):
Looks all good now! 💯 

Also made sure tests were green for net48:
image

awakecoding:
> Looks all good now! 💯
> 
> Also made sure tests were green for net48: image

win-win-x64 in the paths in your screenshot, is that normal?

thenextman:
> > Looks all good now! 💯
> > Also made sure tests were green for net48: image
> 
> win-win-x64 in the paths in your screenshot, is that normal?

How are the paths dealt with in the test project? I ran the test locally and just dropped the native DLL in the `bin\Debug` folder. Is it able to resolve the DLL in a subdirectory?? Maybe with dotnet 6...

CBenoit:
> win-win-x64 in the paths in your screenshot, is that normal?

I believe so, yes, because I specify --arch `win-x64`. This gives `<os>-<arch>`.

CBenoit:
> How are the paths dealt with in the test project? I ran the test locally and just dropped the native DLL in the `bin\Debug` folder. Is it able to resolve the DLL in a subdirectory?? Maybe with dotnet 6...

You have to put the DLLs in the `dependencies` folder at the root of the project. Example for Windows x64: `/dependencies/runtimes/win-x64/native/libDevolutionsPicky.so`. See [here](https://github.com/Devolutions/picky-rs/blob/b1b06ef819500abfe68a391156ff6c1115596863/ffi/dotnet/Devolutions.Picky/Devolutions.Picky.csproj#L17).

PR #185, opened 2022-10-21
feat(dotnet): support returning Pem data

CBenoit:
Until array support is added to diplomat, we do the FFI dance manually.

### DIFF
--- a/ffi/dotnet/Devolutions.Picky/src/Pem.Addons.cs
+++ b/ffi/dotnet/Devolutions.Picky/src/Pem.Addons.cs
@@ -1,0 +1,39 @@
+using System;
+using System.Runtime.InteropServices;
+
+using Devolutions.Picky.Diplomat;
+
+namespace Devolutions.Picky;
+
+public partial class Pem
+{
+	// TODO: maybe this should be part of the Diplomat namespace in DiplomatRuntime.cs
+#if __IOS__
+    private const string NativeLib = "libDevolutionsPicky.framework/libDevolutionsPicky";
+#else
+    private const string NativeLib = "DevolutionsPicky";
+#endif
+
+	/// Returned data should not be modified!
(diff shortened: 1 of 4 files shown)


ffi/dotnet/Devolutions.Picky/src/Pem.Addons.cs
```diff
@@ -0,0 +1,39 @@
...
+	[DllImport(NativeLib, CallingConvention = CallingConvention.Cdecl, EntryPoint = "Pem_peek_data", ExactSpelling = true)]
+	internal static unsafe extern IntPtr PeekData(Raw.Pem* self, out nuint len);
+
+    public byte[] ToData()
+    {
+        unsafe
+        {
+            if (_inner == null)
+            {
+                throw new ObjectDisposedException("Pem");
```
thenextman, line 27: nit: `null` doesn't mean the object was disposed; maybe `NullReferenceException` or `ArgumentException` is more appropriate?

ffi/dotnet/Devolutions.Picky/src/Pem.Addons.cs
```diff
@@ -0,0 +1,39 @@
+using System;
+using System.Runtime.InteropServices;
+
+using Devolutions.Picky.Diplomat;
+
+namespace Devolutions.Picky;
+
+public partial class Pem
+{
+	// TODO: maybe this should be part of the Diplomat namespace in DiplomatRuntime.cs
```
thenextman, line 10: Indeed I've done the some in my stuff; it saves the repetition

thenextman (approved):
LGTM

ffi/dotnet/Devolutions.Picky/src/Pem.Addons.cs
CBenoit: Indeed! But, in this case it does, `_inner` is `null` only when `Pem` has been disposed of (see `Generated/Pem.cs`)

thenextman: Ah ok, great!

PR #190, opened 2022-11-11
iOS net7.0 nuget target

thenextman:
The existing package multi-targets `netstandard2.0` and `Xamarin.iOS10`. The Xamarin target is necessary to define `__IOS__` and decorate certain callbacks with the necessary attributes for AOT compilation.

`Xamarin.iOS` libraries are not forward compatible with applications built using the new `net7.0-ios` target.

This PR splits the managed build into two build targets. When building the regular package, both targets are included - the package is the same as it was before.

Optionally, we can override the package name at build time. If building the `Devolutions.Picky.iOS` target, only the iOS build target is included and we conditionally set the target framework to `net7.0-ios`. Additionally, the ".props" file deployed by the nuget is renamed (it must match the name of the package).

This workaround is necessary until consumers consolidate on net6/7 targets. In the future we will be able to consolidate back to a single package (and using only `dotnet build` instead of `msbuild`).

Additionally, I updated the deployed .targets file to better check the current platform. When installing Devolutions.Picky in an Android project targeting `net7.0-android` on a macOS machine, the build unexpectedly picks up the macOS native reference (this issue already existed, but only becomes a problem on the new target framework(s)).


awakecoding (approved):

CBenoit (approved):
LGTM

PR #208, opened 2023-02-28
Replace ring ecdsa implementation with rust cryptos

Xynnn007:
Hi, I'm working on the downstream crates and find that the `ring` crate used in this crate cannot be compiled in arch `s390x`. Thus I summit this PR to replace the `ring` with RustCrypto to implement Elliptic Curve Signatures.

Please take a look if I did anything contrary to the design of this module. Thanks!

#### Main change
- Replaced `ring` elliptic curve signature implementations with RustCrypto
- Added a compability test for previous ring implementation and current rust crypto to ensure the functionality of API does not change

cc @flavio who is the initial writer of the related code

### DIFF
--- a/picky/src/key/ec.rs
+++ b/picky/src/key/ec.rs
@@ -2,9 +2,8 @@ use crate::key::{KeyError, PrivateKey, PublicKey};
 use picky_asn1::wrapper::{BitStringAsn1, OctetStringAsn1Container};
 use picky_asn1_x509::{oids, private_key_info};
 
-pub(crate) struct EcdsaKeypair<'a> {
+pub(crate) struct EcdsaKeypair {
     pub(crate) private_key: Vec<u8>,
-    pub(crate) public_key: &'a [u8],
     pub(crate) curve: EcdsaCurve,
 }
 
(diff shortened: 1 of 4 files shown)


Xynnn007:
Update: delete `ecdsa` crate. I have test successfully locally and now it is ready to review

Xynnn007:
Update: delete `ecdsa` crate. I have test successfully locally and now it is ready to review

picky/src/key/ec.rs
```diff
@@ -4,7 +4,7 @@ use picky_asn1_x509::{oids, private_key_info};
 
 pub(crate) struct EcdsaKeypair<'a> {
     pub(crate) private_key: Vec<u8>,
-    pub(crate) public_key: &'a [u8],
+    pub(crate) _public_key: &'a [u8],
```
flavio, line 7: I think we can remove this one

picky/src/key/ec.rs
```diff
@@ -65,13 +65,13 @@ impl<'a> TryFrom<&'a PrivateKey> for EcdsaKeypair<'a> {
...
             private_key: private_key_data,
-            public_key: public_key_data,
+            _public_key: public_key_data,
             curve,
         })
     }
 }
 pub(crate) struct EcdsaPublicKey<'a> {
-    pub(crate) data: &'a [u8],
+    pub(crate) _data: &'a [u8],
```
flavio, line 74: I got confused by the `_` prefix, I thought this `_data` attribute was no longer used. Shouldn't we leave the name unchanged?

flavio (commented):
Overall LGTM. I think this would be a nice change to the library. When I wrote the initial code there was no p384 support outside of `ring`.
I'm glad we can switch to a pure rust implementation now :partying_face:

picky/src/key/ec.rs
Xynnn007: Yes, I'll fix this. the `_` was left during coding by mistake

picky/src/key/ec.rs
Xynnn007: This field is now used but only used in test `sign_and_verify_compatibility_with_ring`

flavio: If I remember correctly, the new crypto libraries you're using allow to obtain the public key starting from the private key. I would remove the `_public_key` attribute entirely and change the unit test to obtain the public key starting from the private one.

I hope I explained myself properly :sweat_smile:

Xynnn007: Yea, I understood your suggestion, very clearly! I once thought that the `public key` should be read using `crate::key::ec::EcdsaKeypair::try_from` because it is provided in this crate. Now I changed the way to read the PEM-encoded key using `p256`/`p384` APIs as your suggestions. In this way the API for `crate::key::ec::EcdsaKeypair::try_from` will not be tested. Hope that it is not an important problem

flavio: I like the change.

CBenoit (approved):
Thank you @Xynnn007 , I’m very happy with this change.
And thanks to you too @flavio, your review is really appreciated!
This is looking good to me :100:

Xynnn007:
Thanks for adopting this! @CBenoit When could this PR be included in the next release?

CBenoit:
> Thanks for adopting this! @CBenoit When could this PR be included in the next release?

I’ll take some time to publish a new version next week

PR #209, opened 2023-03-04
ASN.1 Integer minimal encoding is required per X.690-0207 Section 8.3.2

brimworks:
Validation of ASN1.Integer in Go code was failing due to the encoding not being minimal, and the spec indicates that it is required to be minimal.

### DIFF
--- a/picky-asn1/src/wrapper.rs
+++ b/picky-asn1/src/wrapper.rs
@@ -338,6 +338,21 @@ pub struct IntegerAsn1(#[serde(with = "serde_bytes")] pub Vec<u8>);
 
 impls! { IntegerAsn1(VecU8), Tag::INTEGER }
 
+// X.690-0207 Section 8.3.2:
+fn minimal_encode_start(bytes: &[u8]) -> usize {
+    let mut start = 0;
+    while start + 1 < bytes.len() {
+        if bytes[start] == 0 && (bytes[start + 1] & 0x80) == 0 {
+            start = start + 1;
+        } else if bytes[start] == 0xFF && (bytes[start + 1] & 0x80) == 0x80 {
+            start = start + 1;
+        } else {
+            break;
+        }
+    }
+    start
+}
+
 impl IntegerAsn1 {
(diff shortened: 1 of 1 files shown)


CBenoit:
Thank you! I approved CI run. I’ll take more time for reviewing next Monday :smiley:

picky-asn1/src/wrapper.rs
```diff
@@ -379,8 +394,14 @@ impl IntegerAsn1 {
         }
     }
 
+    // TODO: This method should probably take a `&[u8]`:
```
CBenoit, line 397: We can’t take owership of bytes anymore in this case though. Is `start == 0` not common enough? Also, do you have code in mind requiring keeping the `bytes` at call site? In my experience the input object is never used again once the IntegerAsn1 struct is built.

brimworks:
@CBenoit how would you like me to proceed? Would you like me to remove the comment in regard to taking a `&[u8]`?

CBenoit:
> @CBenoit how would you like me to proceed? Would you like me to remove the comment in regard to taking a `&[u8]`?

I’m open to changing the API to a `&[u8]` if you want to.

brimworks:
> > @CBenoit how would you like me to proceed? Would you like me to remove the comment in regard to taking a `&[u8]`?
> 
> I’m open to changing the API to a `&[u8]` if you want to.

How about if I just remove this comment for now, that idea can always be addressed in a separate PR.

CBenoit:
> > > @CBenoit how would you like me to proceed? Would you like me to remove the comment in regard to taking a `&[u8]`?
> > 
> > 
> > I’m open to changing the API to a `&[u8]` if you want to.
> 
> How about if I just remove this comment for now, that idea can always be addressed in a separate PR.

Sounds good!

CBenoit (approved):
Looks good to me! Thank you :smile: 
I’ll publish next candidate version next Monday!

brimworks:
Thanks!

PR #211, opened 2023-04-06
Serialize and deserialize RSASSA-PSS AlgorithmIdentifier.

brimworks:
Note that the deserializer assumes all fields are specified in the ASN.1 structure and none are using the default. Given that SHA1 is considered broken, and that all default fields only make sense in the context of SHA1, no effort has been made to support the DEFAULTs.


CBenoit (approved):
Sorry for the delay! This looks good to me, thanks! :smile:

PR #216, opened 2023-04-27
Update base64 to version 0.21.0

### LINKED ISSUE #215: Cannot build picky-asn1-der and picky-asn1-x509 on Fedora >= 38
When using only the system available dependencies, it is not possible to build `picky-asn1-x509` and `picky-asn1-der` because the only `base64` version available is `0.21`. The Fedora packaging guidelines for Rust packages allow only system available crates to be used when building packages.

Please update `base64` to version `0.21` to unblock updates for packages that depend on `picky-asn1-x509` and `picky-asn1-der` on Fedora.

ansasaki:
* Modifiy the code to adjust the incompatible API changes
* Replace the `GARBAGE` string in pem tests with the corresponding base64 encoded string `R0FSQkFHRQo=`

Fixes #215

### DIFF
--- a/picky/src/ssh/mod.rs
+++ b/picky/src/ssh/mod.rs
@@ -8,13 +8,11 @@ pub use certificate::{SshCertKeyType, SshCertType, SshCertificate, SshCertificat
 pub use private_key::SshPrivateKey;
 pub use public_key::SshPublicKey;
 
-use base64::read::DecoderReader;
-use base64::write::EncoderWriter;
 use byteorder::ReadBytesExt;
 use std::io::{self, Read};
 
-pub type Base64Writer<T> = EncoderWriter<T>;
-pub type Base64Reader<'a, R> = DecoderReader<'a, R>;
+pub(crate) type Base64Writer<'a, T, E> = base64::write::EncoderWriter<'a, T, E>;
+pub(crate) type Base64Reader<'a, T, E> = base64::read::DecoderReader<'a, T, E>;
 
 const SSH_RSA_KEY_TYPE: &str = "ssh-rsa";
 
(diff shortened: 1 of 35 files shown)


picky/src/ssh/mod.rs
```diff
@@ -8,13 +8,11 @@ pub use certificate::{SshCertKeyType, SshCertType, SshCertificate, SshCertificat
...
 
-use base64::read::DecoderReader;
-use base64::write::EncoderWriter;
 use byteorder::ReadBytesExt;
 use std::io::{self, Read};
 
-pub type Base64Writer<T> = EncoderWriter<T>;
-pub type Base64Reader<'a, R> = DecoderReader<'a, R>;
+pub type Base64Writer<'a, T, E> = base64::write::EncoderWriter<'a, T, E>;
+pub type Base64Reader<'a, T, E> = base64::read::DecoderReader<'a, T, E>;
```
CBenoit, line 15: I know this wasn’t originally your code, but do you think you can make these public at crate-level only?
I feel this should not be exposed as part of the public API of picky

```suggestion
pub(crate) type Base64Writer<'a, T, E> = base64::write::EncoderWriter<'a, T, E>;
pub(crate) type Base64Reader<'a, T, E> = base64::read::DecoderReader<'a, T, E>;
```

CBenoit (commented):
LGTM, thank you! I have a minor comment, do you think you could address it?

Cargo.lock in [ffi/wasm](https://github.com/Devolutions/picky-rs/tree/master/ffi/wasm) folder should also be updated.

picky/src/ssh/mod.rs
ansasaki: Sure, I'll fix this

CBenoit (approved):
Looks good! Thank you!

PR #223, opened 2023-06-06
Implemented ECDH-ES algorithm for JWE

### LINKED ISSUE #221: Elliptic curves JWE encryption support (x25519/x448 curves)
Currently,  x25519 and x448 curves are not supported either by picky nor picky-asn1-x509 crates. In scope of this task the following should be implemented:

- PKCS1(DER) and PKCS8 pem keys support
- JWS encryption (ECDH-ES)

**Note:** x509 certs signing could be left unimplemented for x25519 and x448 (ed25519 and ed448 are used instead)

Implementation notes:
The following crates could be used to implement ed* arithmetic:
- [dalek-cryptography/ed25519-dalek](https://github.com/dalek-cryptography/ed25519-dalek)
- [crate-crypto/Ed448-Goldilocks](https://github.com/crate-crypto/Ed448-Goldilocks) - However this crate is very experimental and maybe should be disabled by default by cargo feature

pacmancoder:
### Description
This PR adds `ECDH-ES` algorithm to jwe. From the user standpoint, it does not differ a lot from, for example, RSA JWE encryption. However key generation and cryptographic ops for `ECDH-ES` is much faster.
- The algorithm involves 2 keypairs -- one static, and one ephemeral (encryption private key is discarded right after encryption).
- Both direct encryption via ECDH key and AES key wrapping are implemented. 

I believe Picky now will become the first pure-Rust library that supports all ECDH JWE features and [RFC 8037](https://www.rfc-editor.org/rfc/rfc8037) (X25519 keys)

### Fixes
- Added missing zeroing code for JWE encryption logic
- Fixed `Jwk::from_public_key` for ED keys

Closes #221


picky/src/jose/jwe.rs
```diff
@@ -455,58 +622,78 @@ enum EncoderMode<'a> {
...
                 return Err(JweError::InvalidSize {
                     ty: "symmetric key",
-                    expected: header.enc.key_size(),
+                    expected: jwe.header.enc.key_size(),
                     got: symmetric_key.len(),
                 });
             }
 
             // Override `alg` header with "dir"
-            header.alg = JweAlg::Direct;
```
pacmancoder, line 473: I found and fixed a bug here -- we were setting JweAlg::Direct for direct encryption, but that value was never serialized as the header was serializing to base64 in the code above.

picky/src/jose/jwe.rs
```diff
@@ -455,58 +622,78 @@ enum EncoderMode<'a> {
...
+                    JweAlg::RsaOaep => PaddingScheme::new_oaep::<sha1::Sha1>(),
+                    JweAlg::RsaOaep256 => PaddingScheme::new_oaep::<sha2::Sha256>(),
+                    unsupported => {
+                        return Err(JweError::UnsupportedAlgorithm {
+                            algorithm: format!("{:?}", unsupported),
+                        })
+                    }
+                };
+
+                let cek = generate_cek(jwe.header.enc);
```
pacmancoder, line 658: Moved symmetric key generation to one place, and now I maximized use of `zeroize::Zeroizing` to clear keys on drop

picky/src/jose/jwk.rs
```diff
@@ -30,10 +31,13 @@ pub enum JwkError {
     UnsupportedAlgorithm { algorithm: &'static str },
 
     #[error("Invalid ec public key: {cause}")]
-    InvalidEcPublicKey { cause: &'static str },
+    InvalidEcPublicKey { cause: String },
```
pacmancoder, line 34: `&'static str` is not enough in most cases fro `cause`/`source`

picky/src/jose/jwk.rs
```diff
@@ -371,9 +375,22 @@ impl Jwk {
                     }),
                 }
             }
-            SerdePublicKey::Ed(_) => Err(JwkError::UnsupportedAlgorithm {
-                algorithm: "edwards curves",
-            }),
+            SerdePublicKey::Ed(_) => {
```
pacmancoder, line 378: *Sigh*... Missed it in the previous PR. Now it is covered by ECDH tests

picky/src/jose/jwe.rs
```diff
@@ -168,6 +201,112 @@ pub enum JweAlg {
...
+    /// Decrypts wrapped CEK using the given AES decryption key
+    ///
+    /// SAFETY:
+    ///   - Caller must unsure `decryption_key` size matches the wrapping algorithm
+    fn decrypt_key(
+        &self,
+        cek_alg: JweEnc,
+        encrypted_cek: &[u8],
+        decryption_key: &[u8],
+    ) -> Result<Zeroizing<Vec<u8>>, JweError> {
```
CBenoit, line 229: I guess this function is not really "unsafe" like in "memory safety".
Do you think you could clarify what happens when `decryption_key` size does not matches the wrapping algorithm? Is it returning an `Err`? Does it cause a panic? A misbehavior?

In any cases, "Safety" sections are typically used to document invariants that must be uphold when calling an unsafe function.

For reference, the [API guidelines book](https://rust-lang.github.io/api-guidelines/documentation.html#function-docs-include-error-panic-and-safety-considerations-c-failure) says:

> Unsafe functions should be documented with a "Safety" section that explains all invariants that the caller is responsible for upholding to use the function correctly.

Maybe calling this section "Invariant" can help clarify to the reader that there is no hidden "memory safety"  considerations to have with this function:
```suggestion
    /// Decrypts wrapped CEK using the given AES decryption key
    ///
    /// # Invariants
    ///
    /// - Caller must unsure `decryption_key` size matches the wrapping algorithm
    fn decrypt_key(
        &self,
        cek_alg: JweEnc,
        encrypted_cek: &[u8],
        decryption_key: &[u8],
    ) -> Result<Zeroizing<Vec<u8>>, JweError> {
```

picky/src/jose/jwk.rs
CBenoit: Thank you!

CBenoit (commented):
As always, thank you for the well-documented high quality code!

picky/src/jose/jwe.rs
```diff
@@ -22,6 +26,13 @@ type Aes192Gcm = aes_gcm::AesGcm<aes_gcm::aes::Aes192, aes_gcm::aead::generic_ar
 #[derive(Debug, Error)]
 #[non_exhaustive]
 pub enum JweError {
+    /// JWK conversion error
+    #[error("JWK conversion error: {source}")]
```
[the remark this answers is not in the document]
pacmancoder: Yea, I agree, I was thinking about the current error-reporting situation in picky too, I kept using the source's Display in the error description for the sake of consistency, but I agree, I'll fix this for newly added errors, and I think we should create issue and fix this for all existing picky errors with source field (well, it is breaking change, but bill greatly improve library error reporting).

picky/src/jose/jwe.rs
pacmancoder: Well, it will panic (`GenericArray` will throw it). Agreed, `Invariant` will sound better!

pacmancoder: done

picky/src/jose/jwe.rs
```diff
@@ -64,6 +82,15 @@ pub enum JweError {
         expected: usize,
         got: usize,
     },
+
+    #[error("Private and public key algorithms don't match: {context}")]
```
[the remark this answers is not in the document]
pacmancoder: done

picky/src/jose/jwe.rs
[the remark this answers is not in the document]
pacmancoder: done

picky/src/jose/jwe.rs
```diff
@@ -30,6 +41,13 @@ pub enum JweError {
     #[error("AES-GCM error (opaque)")]
     AesGcm,
 
+    /// AES-KW error
+    #[error("AES-KW error: {context}")]
+    AesKw {
+        // `aes_kw::Error` does not implement `std::error::Error`, so we can't use `source` here
```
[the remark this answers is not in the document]
pacmancoder: done

ffi/wasm/Cargo.lock
```diff
@@ -23,20 +23,40 @@ dependencies = [
...
+
 [[package]]
 name = "aes-gcm"
 version = "0.9.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "df5f85a83a7d8b0442b6aa7b504b8212c1733da07b98aae43d4bc21b2cb3cdf6"
 dependencies = [
  "aead",
- "aes",
+ "aes 0.7.5",
```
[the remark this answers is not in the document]
pacmancoder: Updated `aes-gcm` to `0.10`, now `aes` duplication is gone

CBenoit (approved):
LGTM!

picky/src/jose/jwe.rs
CBenoit: Oh, thank you!

Issue #217, opened 2023-05-01
Support pfx files (PKCS#12)


CBenoit:
JIRA Issue: [ARC-53](https://devolutions.atlassian.net/browse/ARC-53)

[ARC-53]: https://devolutions.atlassian.net/browse/ARC-53?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #224, opened 2023-06-16
PKCS12 ASN.1 structures support (picky-asn1-x509)

pacmancoder:
### Description
This PR adds PKCS12 ASN.1 structures serialization/deserialization to `picky-asn1-x509` crate

The main ideas around PKCS12 module implementation (copied from `pkcs12` module docs)
```rust
//! Direct ASN.1 type wrappers interaction is hidden inside serialiation/deserialization logic where
//! it is possible and makes sense. For example `SEQUENCE` is represented as `Vec<T>` in struct
//! definition. However, most basic types such as `OCETET STRING` or `BMPString` are represented by
//! their own distinctive types to avoid confusion.
//!
//! Structures defined in this module are were built in such wat that allows maximum backward and
//! forwards compatibility with other PKCS#12 implementations by wrapping unknown variants and
//! algorithms inside raw ASN1 types. For example, if the client code knows about specific
//! PKCS12 attribute, it always could be accessed via matching on attribute OID and raw ASN1
//! attribute value.
//!
//! `SafeBag` contents are presented as `OCTET STRING` or `Asb1RawDer` to allow more flexibility
//! when reading/writing PFX contents. For example, even if some specific `PrivateKeyInfo` or
//! `Certificate` is not supported by `picky` directly, they still could be read/written as raw
//! data. See tests in this file for full unencrypted PFX file contents parsing.
//!
//! Instead of using more general `AlgorithmIdentifier` and `ContentInfo` types defined in this
//! crate, this module defined pkcs12-specific wrappers with more usable interface, while providing
//! access to raw ASN1 types when needed (not supported structure variant).
//!
//! Parsing code is tested against PFX files generated by `OpenSSL` and `certmgr`
```

### Fixes
- Renamed `oids::pkcs7` to correct `oids::content_info_type_data`

### Note to reviewer
This PR indeed huge. But a lot of the LoC goes from tests, test assets, docs and a lot of boilerplate for serde-based serialization/deserialization code

### References
**RFC:**  https://datatracker.ietf.org/doc/html/rfc7292
**Issue:** Part of #217 implementation


test_assets/pkcs12/leaf_password_is_abc.parsed.bin
```diff
@@ -0,0 +1,60 @@
+Pfx {
```
CBenoit, line 1: Since this is the Debug representation and not actual binary, maybe the extension should be `.txt`. What do you think?

pacmancoder: Yep, I agree

pacmancoder: Done

picky-asn1-x509/src/pkcs12.rs
```diff
@@ -0,0 +1,233 @@
+//! This module implements PKCS#12 specification defined in [RFC7292](https://tools.ietf.org/html/rfc7292).
+//!
+//! Direct ASN.1 type wrappers interaction is hidden inside serialiation/deserialization logic where
+//! it is possible and makes sense. For example `SEQUENCE` is represented as `Vec<T>` in struct
+//! definition. However, most basic types such as `OCETET STRING` or `BMPString` are represented by
+//! their own distinctive types to avoid confusion.
+//!
+//! Structures defined in this module are were built in such way that allows maximum backwards and
```
CBenoit, line 8: Typo?
```suggestion
//! Structures defined in this module are built in such a way that allows maximum backwards and
```

picky-asn1-x509/src/pkcs12.rs
```diff
@@ -0,0 +1,233 @@
+//! This module implements PKCS#12 specification defined in [RFC7292](https://tools.ietf.org/html/rfc7292).
```
CBenoit, line 1: I believe the `mod.rs` convention is used everywhere else in picky codebase, so I suggest we stick to it.

As an aside, I personally prefer the `mod.rs` convention so that all files related to a given module are all in the same folder, but I’m willing to be convinced.

picky-asn1-x509/src/pkcs12.rs
```diff
@@ -0,0 +1,233 @@
...
+        // This test validates PFX structure parsing and parsing of `Certificate`/`PrivateKeyInfo`
+        // structures encapsulated inside PFX
+
+        let encoded = include_bytes!("../../test_assets/pkcs12/openssl_nocrypt.pfx");
+        let decoded: Pfx = picky_asn1_der::from_bytes(encoded).unwrap();
+
+        // Validate parsed PFX structure via debug representation
+        expect_file!["../../test_assets/pkcs12/openssl_nocrypt.parsed.bin"].assert_debug_eq(&decoded);
+
+        // OpenSSL-generate PFXs will have exactly the same bytes when re-encoded
```
CBenoit, line 133: Typo?
```suggestion
        // OpenSSL-generated PFXs will have exactly the same bytes when re-encoded
```

picky-asn1-x509/src/algorithm_identifier.rs
```diff
@@ -1083,4 +1172,22 @@ mod tests {
...
+            picky_asn1_der::from_bytes(&encoded).expect("failed RawAlgorithmIdentifier deserialization");
+
+        let expected = RawAlgorithmIdentifier::from_parts(oids::hmac_with_sha256(), None);
+
+        pretty_assertions::assert_eq!(decoded, expected);
+
+        check_serde!(decoded: RawAlgorithmIdentifier in encoded);
+    }
+
+    // TODO PACMANCODER: Add test with parameters set to some value
```
CBenoit, line 1192: Nit: it’s frequent to attach TODO comment author this way:
```suggestion
    // TODO(pacmancoder): Add test with parameters set to some value
```

picky-asn1-x509/src/pkcs12/authenticated_safe.rs
```diff
@@ -0,0 +1,102 @@
...
+                let mut seq = serializer.serialize_seq(Some(2))?;
+                let oid: ObjectIdentifierAsn1 = oids::content_info_type_data().into();
+                seq.serialize_element(&oid)?;
+                let content: ExplicitContextTag0<OctetStringAsn1Container<Asn1SequenceOf<SafeContentsContentInfo>>> =
+                    OctetStringAsn1Container(Asn1SequenceOf(safe_contents.clone())).into();
+                seq.serialize_element(&content)?;
+                seq.end()
+            }
+            AuthenticatedSafeContentInfo::Unknown { content_type, content } => {
+                let sequence_length = 1 + content.is_some() as usize;
```
CBenoit, line 39: Nit: it is generally well-advised to not use "as" casting in most cases.

`From` is only implemented when the conversion is lossless and infallible. It’s incredibly useful to have a compile-time error when code is changed later.

When conversion is fallible, let’s use `TryFrom`, it’s fine to just `usize::try_from(v).unwrap()` in most cases. This way, it’s simpler to audit fallible conversions using `grep` and it’s easier to debug (we’ll get an immediate crash along a stacktrace when the value is invalid). Of course, in production code, we may want to handle the error in a more elegant manner, but if we can be sure it never fails, it’s not strictly necessary (in this case, the conversation can never fail because `usize` can always hold values 0 and 1).

Note that `usize::try_from(v).unwrap()` does not incur any performance penalty when redundant. For instance, casting a u32 to usize is actually infallible on 32 and 64 bits architectures and code is optimized accordingly: https://gcc.godbolt.org/z/Y4WjWTqh1
On the other hand, we actually want to check when this property does not actually hold.

For reference, I consider using `as` to be appropriate when:
- Using it as a building block to implement `From` or `Into`
- Using it as a building block to implement `as_X` methods

Clippy lint about this: https://rust-lang.github.io/rust-clippy/master/#cast_lossless

Nit 2: this pattern is not very frequent, and it took me some time to realize this is basically equivalent to the following:
```rust
let sequence_length = content.is_some() {
    2
} else {
    1
};
```
However, this sounds useful when there are a lot of optional elements. If it appears at other places, it’s good to stick with the `usize::from` conversion instead of the `if … else …` shape.

picky-asn1-x509/src/pkcs12/encryption/pbes2/pbkdf2.rs
```diff
@@ -0,0 +1,217 @@
...
+}
+
+impl ser::Serialize for Pbkdf2Params {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: ser::Serializer,
+    {
+        use ser::SerializeSeq;
+
+        let sequence_size = 2 + self.key_length.is_some() as usize + self.prf.is_some() as usize;
```
CBenoit, line 41: Nit: speaking of which! Here I simply suggest we use `usize::from` instead of the `as` as mentioned above. If you decide to address this nit, can you do the same at all the other places? Thank you!

picky-asn1-x509/src/pkcs12/encryption/pbes1.rs
```diff
@@ -0,0 +1,37 @@
...
+/// [PKCS #12: Personal Information Exchange Syntax Standard Version](https://tools.ietf.org/html/rfc7292#appendix-C)
+/// # Appendix C. Keys and IVs for Password Privacy Mode
+/// ```not_rust
+/// pkcs-12PbeParams ::= SEQUENCE {
+///     salt        OCTET STRING,
+///     iterations  INTEGER
+/// }
+/// ```
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
```
CBenoit, line 33: Nit: superfluous newline between the doc comment and the actual function

picky-asn1-x509/src/pkcs12/safe_bag.rs
```diff
@@ -0,0 +1,455 @@
...
+use serde::{de, ser};
+
+pub use certificate::*;
+pub use crl::*;
+pub use key::*;
+pub use secret::*;
+
+/// [PKCS #12: Personal Information Exchange Syntax Standard](https://datatracker.ietf.org/doc/html/rfc7292#section-4.2)
+///
+/// SafeBag is the atomic node of PKCS#12 structure, it could contain concrete data such as
```
CBenoit, line 21: Nit: it’s not exactly clear to me what atomic means in this situation. Is my understanding correct?
```suggestion
/// SafeBag is the smallest unit of information in PKCS#12 SafeContents, it could contain concrete data such as
```
Though I’m not sure we can say it’s the smallest unit of information since a `SafeBag` can contain other `SafeBag`s recursively.
```suggestion
/// SafeBag a building block of PKCS#12 structure, it could contain concrete data such as
```

picky-asn1-x509/src/pkcs12/safe_bag/certificate.rs
```diff
@@ -0,0 +1,106 @@
...
+///     -- DER-encoded X.509 certificate stored in OCTET STRING
+/// ...
+/// CertTypes BAG-TYPE ::= {
+///     x509Certificate |
+///     sdsiCertificate,
+///     ... -- For future extensions
+/// }
+/// ```
+
+#[derive(Debug, Clone, PartialEq, Eq)]
```
CBenoit, line 27: Nit: newline

CBenoit (commented):
Code is well documented, tested and written in a straightforward manner. I like the simplicity of `expect-test`-driven tests.
Only a few minor comments below, good to merge once this is all addressed!

picky-asn1-x509/src/pkcs12.rs
pacmancoder: yep, definitely, thanks!

picky-asn1-x509/src/pkcs12.rs
pacmancoder: Specifically in `picky-asn1-x509` there we only single example of module subfolder - `pkcs7` and it does not use `mod.rs` convention, so I used same approach in `pkcs12` but I completely agree with you, mod.rs is definitely _superior_ 😄 

Well, then while I am at it, I'll restructure `pkcs7` too

picky-asn1-x509/src/pkcs12/authenticated_safe.rs
pacmancoder: `bool` -> `usize` conversion is infallible and well-defined (`false` -> `0`, `true` -> `1`) and in that case, it is just more compact alternative to `usize::from(val)`, and it has relatively the same level of readability as `usize::from`. This pattern is used in lot of places for optional values in `pkcs12` code, and it is much more compact than `if ... {} else ...`, especially in cases where there is more than one optional value:
```
let sequence_size = 2 + self.key_length.is_some() as usize + self.prf.is_some() as usize;
```

picky-asn1-x509/src/pkcs12/safe_bag.rs
pacmancoder: Yeah, `a building block` sounds right

picky-asn1-x509/src/pkcs12.rs
CBenoit: Oh, I see! I probably missed that in a previous review, thank you!

picky-asn1-x509/src/pkcs12/safe_bag.rs
```diff
@@ -0,0 +1,455 @@
...
+use serde::{de, ser};
+
+pub use certificate::*;
+pub use crl::*;
+pub use key::*;
+pub use secret::*;
+
+/// [PKCS #12: Personal Information Exchange Syntax Standard](https://datatracker.ietf.org/doc/html/rfc7292#section-4.2)
+///
+/// SafeBag a building block of PKCS#12 structure, it could contain concrete data such as
```
CBenoit, line 21: (Typo from my previous suggestion)
```suggestion
/// SafeBag is a building block of PKCS#12 structure, it could contain concrete data such as
```

CBenoit (approved):
LGTM!

PR #225, opened 2023-07-01
Add support for PKCS#12(PFX) in `picky`

### LINKED ISSUE #217: Support pfx files (PKCS#12)

pacmancoder:
# Description
Adds support for PKCS#12(PFX) files in Picky.

Extracted from `picky/src/pkcs12/mod.rs`:
```
//! # PFX file structure
//! PFX files are constructed from `safe contents` on the highest level, which are containers for
//! various types of `safe bags`. They could be encrypted or not and contain one or more `safe bags`.
//!
//! `Safe bags` are polymorphic containers for various types of data, such as private keys, certificates,
//! custom data (`secrets`), CRLs, etc. They could also be nested, which means that one `safe bag` could
//! contain multiple `safe bags` inside, making it possible to represent complex tree-like structures.
//!
//! Each `safe bag` could also contain `attributes`, which are key-value pairs of various types of data.
//! High-level API defines only `friendly name` and `local key id` attributes, but it is possible to
//! parse/build any custom attributes as well.
//!
//! There are also special `safe bags` called `shrouded key bags`, which are basically encrypted
//! PKCS8 private keys. Encryption on `safe bag` level is performed to allow PFX files that have
//! unencrypted certificates and encrypted private keys, to be preprocessed by the software without
//! knowing PFX password (e.g. check if PFX contains private key for the certificate or not). Usually
//! relations between contained certificates and private keys are marked with `local key id` attribute
//! which is set to the same value for both certificate and private key.
//!
//! # Parsing
//! Almost any kind of PKCS#12 object specified by [RFC](https://datatracker.ietf.org/doc/html/rfc7292)
//! could be parsed or constructed. However, only the most used PFX types are wrapped in high level
//! convenient API such as private keys, certificates, nested pfx nodes, etc. If parsed
//! data was in fact some rarely used PFX type, it is still possible to access it via low level API
//! via `.inner()` method usually returns ASN.1 DER encoded data (`picky-asn1-x509` crate types).
//!
//! Parsing process could be controlled even more with [`Pkcs12ParsingParams`], which allows to skip
//! some parsing errors (e.g. Skip failed mapping of private key ASN.1 structure to high-level picky wrapper)
//!
//! Parsing of PFX file is simple as calling [`Pfx::from_der_bytes`] method with required params
//! (crypto context which holds PFX password, parsing params and data itself), after which its safe
//! bags could be inspected and data extracted from PFX.
//!
//! # Building
//! Building process of PFX files is down-to-top, in contrast to parsing. It starts with
//! building of required safe bags/attributes structures, wrapping them inside safe contents
//! (encrypted or unencrypted) and then wrapping them in PFX structure. This allows to keep API
//! flexible while representing Pfx file both after parsing and building as the same type ([`Pfx])
//!
//! # Encryption
//! - It is advised to always use Pbes2 AES-based encryption for PFX files. The only use case for new
//! PFX files with Pbes1 encryption is to support legacy software that does not support Pbes2.
//! - Usually PFX files without passwords are actually encrypted with empty-string passwords if
//! generated by modern software by default. (e.g. certmgr or OpenSSL). However, it is possible to
//! create PFX files without any encryption and MAC in picky if PFX will be used just as a plain
//! container for certificates and private keys and stored somewhere securely (e.g. wrapped in
//! another encryption layer).
```

Closes  #217


CBenoit (commented):
Thank you! I’ll continue reviewing later today. Here is a few comments I got so far.

picky/src/checked_string.rs
```diff
@@ -0,0 +1,124 @@
...
+
+        Ok(Self(bmp_string))
+    }
+}
+
+/// Wrapper around `picky-asn1-x509`'s `BMPString` type which provides convenient methods for
+/// from/into UCS-2 encding of `BMPString`.
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize)]
+#[serde(transparent)]
+pub struct CheckedBmpString(Asn1BmpString);
```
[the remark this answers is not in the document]
pacmancoder: Agreed, nice idea!

Cargo.lock
```diff
@@ -2491,13 +2609,13 @@ version = "0.42.0"
...
- "windows_x86_64_gnu",
- "windows_x86_64_gnullvm",
- "windows_x86_64_msvc",
+ "windows_aarch64_gnullvm 0.42.1",
+ "windows_aarch64_msvc 0.42.1",
+ "windows_i686_gnu 0.42.1",
+ "windows_i686_msvc 0.42.1",
+ "windows_x86_64_gnu 0.42.1",
+ "windows_x86_64_gnullvm 0.42.1",
+ "windows_x86_64_msvc 0.42.1",
```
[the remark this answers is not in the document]
pacmancoder: Yeah, I'll look into that

picky/src/pkcs12/mod.rs
```diff
@@ -0,0 +1,606 @@
...
+                    mac_data: self.mac_data.as_ref().map(|mac_data| mac_data.inner().clone()),
+                };
+
+                picky_asn1_der::to_vec(&pfx_asn1).map_err(Into::into)
+            }
+        }
+    }
+
+    /// Parse PFX file from DER bytes
+    pub fn from_der_bytes(
```
CBenoit, line 154: Nit: I think other similar methods are simply named `from_der`.

picky/src/pkcs12/mod.rs
```diff
@@ -0,0 +1,606 @@
...
+    pub fn new_without_mac(safe_contents: Vec<SafeContents>) -> Self {
+        Self {
+            safe_contents,
+            mac_data: None,
+            auth_safe_data: None,
+        }
+    }
+
+    /// Serialize PFX file to DER bytes
+    pub fn to_der_bytes(&self) -> Result<Vec<u8>, Pkcs12Error> {
```
CBenoit, line 125: Nit: similarly, I remember other methods being called `to_der`

picky/src/pkcs12/mod.rs
```diff
@@ -0,0 +1,606 @@
...
+//! could be parsed or constructed. However, only the most used PFX types are wrapped in high level
+//! convenient API such as private keys, certificates, nested pfx nodes, etc. If parsed
+//! data was in fact some rarely used PFX type, it is still possible to access it via low level API
+//! via `.inner()` method usually returns ASN.1 DER encoded data (`picky-asn1-x509` crate types).
+//!
+//! Parsing process could be controlled even more with [`Pkcs12ParsingParams`], which allows to skip
+//! some parsing errors (e.g. Skip failed mapping of private key ASN.1 structure to high-level picky
+//! wrapper)
+//!
+//! Parsing of PFX file is simple as calling [`Pfx::from_der_bytes`] method with required params
```
CBenoit, line 33: (If you address the other comments, you’ll need to update this as well)

picky/src/pkcs12/encryption.rs
```diff
@@ -0,0 +1,688 @@
...
+    Pbkdf2SaltSource as Pbkdf2SaltSourceAsn1, Pkcs12EncryptionAlgorithm as Pkcs12EncryptionAsn1,
+};
+use std::str::FromStr;
+
+/// Same default KDF iterations as in OpenSSL
+const DEFAULT_KDF_ITERATIONS: usize = 2048;
+const DEFAULT_SALT_SIZE: usize = 8;
+const AES_BLOCK_SIZE: usize = 16;
+
+pub trait Pkcs12Rng: rand::RngCore + rand::CryptoRng {}
```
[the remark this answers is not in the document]
pacmancoder: Huh... Looks like i was looking at some old `rand` sources where RngCore bound on CryptoRng was absent... Thanks for spotting that! 👍

picky/src/pkcs12/mac.rs
```diff
@@ -0,0 +1,187 @@
...
+/// Parsed MAC algorithm parameters
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum Pkcs12MacAlgorithm {
+    Hmac(Pkcs12MacAlgorithmHmac),
+    Unknown,
+}
+
+/// Parsed PFX MAC data
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Pkcs12MacData {
```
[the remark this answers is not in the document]
pacmancoder: We actually do need in public API so developer could inspect mac data of the parsed file... But the real issue here is that I dos not add `mac_data()` method to `Pfx`, I will fix that

picky/src/pkcs12/encryption.rs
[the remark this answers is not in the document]
pacmancoder: Aha, I got it, `rand 0.8` does not have `CryptoRng: RngCore` bound, I'll update `rand` to `0.9`

picky/src/pkcs12/mac.rs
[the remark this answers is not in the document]
CBenoit: Sounds good!

picky/src/pkcs12/encryption.rs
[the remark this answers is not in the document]
pacmancoder: As discussed in slack, because rand 0.9 is not released yet, we will keep Pkcs12Rng trait, but it will stripped from the public API

picky/src/pkcs12/mac.rs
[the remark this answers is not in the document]
pacmancoder: Done, now method is available

picky/src/pkcs12/mac.rs
```diff
@@ -0,0 +1,187 @@
...
+
+    pub fn with_iterations(mut self, iterations: u32) -> Self {
+        self.iterations = Some(iterations);
+        self
+    }
+}
+
+/// Parsed MAC algorithm parameters
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum Pkcs12MacAlgorithm {
```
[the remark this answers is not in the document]
pacmancoder: Now used for `Pfx::mac_data`

picky/src/pkcs12/mac.rs
```diff
@@ -0,0 +1,187 @@
...
+pub enum Pkcs12MacError {
+    #[error("Invalid hmac input size")]
+    InvalidHmacInputSize,
+    #[error("MAC validation failed (wrong password or corrupted data)")]
+    MacValidation,
+}
+
+/// HMAC algorithm parameters (used for PFX integrity data)
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Pkcs12MacAlgorithmHmac {
```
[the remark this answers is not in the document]
pacmancoder: Now used for `Pfx::mac_data`

picky/src/pkcs12/mod.rs
pacmancoder: Done

picky/src/pkcs12/mod.rs
pacmancoder: Done

picky/src/pkcs12/mod.rs
pacmancoder: Done

Cargo.lock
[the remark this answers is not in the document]
pacmancoder: Fixed by downgrading dev-dependency `tempfile` to version `3.3`.

picky/Cargo.toml
```diff
@@ -75,7 +75,7 @@ rc2 = { version = "0.8", optional = true }
 pbkdf2 = { version = "0.12", optional = true }
 hmac = { version = "0.12", optional = true }
 
-ucs2 = { version = "0.3", optional = true }
+ucs2 = "0.3"
```
CBenoit, line 78: Should we keep this optional, maybe as a default feature?

picky/Cargo.toml
```diff
@@ -68,23 +68,33 @@ aes-gcm = { version = "0.10", optional = true }
...
 
 [dev-dependencies]
 pretty_assertions = "1.3.0"
 hex = "0.4.3"
 cfg-if = "1.0.0"
 ring = "0.16.20"
 rstest = "0.12.0"
+expect-test = "1"
+tempfile = "=3.3"
```
CBenoit, line 87: Nit: a comment explaining why we are pinning this dependency could help later maintenance

picky/Cargo.toml
pacmancoder: Idk, ucs2 is actually very small crate, around ~200 LoC. It is used by pkcs12 feature internally (for password formatting) but we could keep it as optional feature, I actually don't have a strong opinion on that, I think we could keep it non-optional as it may be useful for other things in picky (such as certificate building), but it is ok if we will keep it optional. What are your thoughts on that?

picky/Cargo.toml
pacmancoder: Yep, agreed, I'll add that

picky/Cargo.toml
CBenoit: Fair enough, let’s keep it!

picky/Cargo.toml
pacmancoder: Done

CBenoit (approved):
LGTM!

PR #226, opened 2023-07-05
Fixed AuthorityKeyIdentifier parsing logic

pacmancoder:
During PKCS12 implementation I found a bug in X509 certificates parsing. It turns out, the `AuthorityKeyIdentifier` attribute parsing logic was incorrect.

Fixed the bug and updated pkcs12 tests


CBenoit (approved):
LGTM!

PR #227, opened 2023-07-26
chore(deps): do not enable default features of chrono

flavio:
Enabling the default features of chrono causes the time 0.1 crate to be added as a transitive dependency.

This old version of time is affected by [CVE RUSTSEC-2020-0071](https://rustsec.org/advisories/RUSTSEC-2020-0071).

Thanks to work done inside of chrono 0.4, there are high chances the majority of the codebases do not actually need it. Which is exactly the case for picky.

### DIFF
--- a/picky-asn1/Cargo.toml
+++ b/picky-asn1/Cargo.toml
@@ -19,7 +19,7 @@ readme = "README.md"
 serde = { version = "1.0.152", default-features = false, features = ["derive"] }
 oid = { version = "0.2.1", default-features = false, features = ["serde_support"] }
 serde_bytes = "0.11.9"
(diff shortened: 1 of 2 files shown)


CBenoit (approved):
Looks good to me! Thank you!

flavio:
Thanks! Do you plan to tag another RC of picky? It would be great to be able to consume this change :smile:

CBenoit:
Yes, I want to publish a new RC version soon especially because of this patch :+1: 
I added this to my schedule for this week.

CBenoit:
@flavio picky.rc-7 that is including this patch is now published on crates.io :tada:

PR #232, opened 2023-08-17
chore: include license files in published crates

### LINKED ISSUE #230: please include license files in published crates
This is a followup for https://github.com/Devolutions/picky-rs/issues/56. It appears that there was a misunderstanding about some things, so I will try to clarify:

Both the MIT and the Apache-2.0 license terms contain a requirement that redistributed sources (i.e. via crates.io, or via Linux distribution packages) contain a copy of the original license text.

So for our purposes (RPM packages for Fedora Linux), it is indeed a requirement to include the original license texts, and the same could be argued for crates distributed via crates.io (it is not a *general* requirement, but it *is* a requirement of the Apache-2.0 and MIT licenses).

I can submit a pull request to add symlinks so that cargo publish includes license files in published crates (full copies of the files are *not* required, unless you're publishing from an operating system that does not support symlinks properly, i.e. Windows).

decathorpe:
I've added symbolic links to the license files to all workspace members that don't have `package.publish = false` set in their Cargo.toml file, and adapted `package.include` settings to include these files where necessary. I've tested the change locally, and `cargo package` included the license texts in `.crate` tarballs as intended.

Fixes #230

### DIFF
--- a/picky-asn1-der/Cargo.toml
+++ b/picky-asn1-der/Cargo.toml
@@ -14,7 +14,7 @@ description = "An ASN.1-DER subset for serde"
 license = "MIT OR Apache-2.0"
 repository = "https://github.com/Devolutions/picky-rs"
 readme = "README.md"
-include = ["src/**/*", "README.md", "CHANGELOG.md"]
+include = ["src/**/*", "README.md", "CHANGELOG.md", "LICENSE-*"]
 
 [dependencies]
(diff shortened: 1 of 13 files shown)


CBenoit (approved):
Looking good to me, thank you!

decathorpe:
Thanks for merging!

PR #233, opened 2023-08-18
picky-krb: CredSSP credentials structures

TheBestTvarynka:
Hello,
During the Kerberos smart card logon, the smart card credentials are transferred using the `TsSmartCardCreds` and `TsCspDataDetail` structures.

The password-based credentials are encoded using the [`ber`](https://github.com/Devolutions/sspi-rs/tree/master/src/ber) module in the `sspi-rs`. But now we have `picky-*` crates for the `asn1` encoding. So, I decided to add CredSSP-related constants and credentials structures to the `picky-krb` crate.

Docs and reference:

* [TSCredentials](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-cssp/94a1ab00-5500-42fd-8d3d-7a84e6c2cf03)
* [TSPasswordCreds](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-cssp/17773cc4-21e9-4a75-a0dd-72706b174fe5)
* [TSSmartCardCreds](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-cssp/4251d165-cf01-4513-a5d8-39ee4a98b7a4)
* [TSCspDataDetail](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-cssp/34ee27b3-5791-43bb-9201-076054b58123)

### DIFF
--- a/picky-krb/src/constants.rs
+++ b/picky-krb/src/constants.rs
@@ -142,6 +142,17 @@ pub mod cksum_types {
     pub const HMAC_SHA1_96_AES256: usize = 16;
 }
 
+//= [2.2.1.2 TSCredentials](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-cssp/94a1ab00-5500-42fd-8d3d-7a84e6c2cf03) =//
+pub mod cred_ssp {
+    pub const TS_PASSWORD_CREDS: u8 = 1;
+    pub const TS_SMART_CARD_CREDS: u8 = 2;
+    pub const TS_REMOTE_GUARD_CREDS: u8 = 6;
+
+    // [KeySpec values](https://learn.microsoft.com/en-us/windows-server/identity/ad-fs/technical-reference/ad-fs-and-keyspec-property#keyspec-values-and-associated-meanings)
+    pub const AT_KEYEXCHANGE: u8 = 1;
+    pub const AT_SIGNATURE: u8 = 2;
+}
+
 pub mod error_codes {
     //= [Error Codes](https://datatracker.ietf.org/doc/html/rfc4120#section-7.5.9) =//
     pub const KDC_ERR_NONE: u32 = 0;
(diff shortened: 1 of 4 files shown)


picky-krb/src/constants.rs
```diff
@@ -141,6 +141,16 @@ pub mod cksum_types {
     pub const HMAC_SHA1_96_AES128: usize = 15;
     pub const HMAC_SHA1_96_AES256: usize = 16;
 }
+//= [2.2.1.2 TSCredentials](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-cssp/94a1ab00-5500-42fd-8d3d-7a84e6c2cf03) =//
```
CBenoit, line 144: Just adding a blank line

```suggestion

//= [2.2.1.2 TSCredentials](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-cssp/94a1ab00-5500-42fd-8d3d-7a84e6c2cf03) =//
```

CBenoit (commented):
This is looking good to me!

I see you authored a rather clean history on your branch so if you want, we can "rebase and merge" the PR, but there is small details I want fixed by rebasing interactively first:
- The commit adding tests should either be part of the commit adding the feature, or alternatively the conventional commits type "test" should be used instead of "feat"
- We don’t really scope commits by module
- We don’t end the summary with `;`
- The summary line should be no longer than 50 characters (and at the very most 72 characters), so that it doesn’t get wrapped onto the next line like here:
image
- [Rebase interactively](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History) my "[suggestion commit](https://github.com/Devolutions/picky-rs/pull/233/commits/fda2fad12a704f2d5008f1ec4df62e5ca8a5e8ff)" into [the relevant commit](https://github.com/Devolutions/picky-rs/pull/233/commits/e78b4d733d6171070f95b8652a78fefd340cd1de) (using the `fixup` rebase command)

This would give the following history:
- feat(picky-krb): add credssp password and smartcard structs
- feat(picky-krb): add creds and key spec constants
- test(picky-krb): unit tests for credssp structures
- feat(picky-asn1-x509): add USER_PRINCIPAL_NAME oid

Following this convention helps me when generating the changelogs.

Otherwise I usually squash the whole PR into a single commit for simplicity.

CBenoit (approved):
Thank you! @TheBestTvarynka 
I’ll rebase and merge into master :smile: 
FYI, I’m wrapping up some other work after what I’ll cut a release for a few picky crates.

PR #234, opened 2023-08-19
ARC-158 - extend the FFI API with PFX and PKCS1 formats

CBenoit:
cc @pacmancoder 

## PKCS12 API

This API enables creation, decoding and inspection of PKCS12 (PFX)
archives. It’s possible to embed any number of X.509 certificates
and private keys. Furthermore, each embedded element may have any number
of attributes attached to it.

## PKCS1 API

This adds the `ToPkcs1` and `FromPkcs1` methods to `PublicKey`.


picky/src/pkcs12/safe_bag.rs
```diff
@@ -20,7 +20,7 @@ pub struct SafeBag {
 }
 
 impl SafeBag {
-    /// Create new safe bag with private key
+    /// Create new safe bag holding a private key
     pub fn new_key(key: PrivateKey, attributes: Vec<Pkcs12Attribute>) -> Result<Self, Pkcs12Error> {
```
pacmancoder, line 24: Maybe we could remove `attributes` from `new_key` after we added `add_attribute`. WDYT?

CBenoit: I wasn’t too sure because `add_attribute` is less efficient (recomputing the inner value with each call), but I can do that yes!

thenextman:
@CBenoit Superficially this all looks good to me; I propose merging and once the new version is available I can start the integration task. I know that encrypted PKCS8 is a follow-up task, so there's still work to do. I'll try to move forward on my side ASAP so I can highlight any problem areas.

thenextman (approved):

Issue #213, opened 2023-04-13
picky-signtool: replace `lief-cwal` crate with `object` crate (pure Rust)


pkking:
Is there a plan to build picky as a pure rust binary/crates as some unsafe code still lives there, e.g. lief-cwal :)

CBenoit:
lief-cwal is a dependency of [picky-signtool](https://github.com/Devolutions/picky-rs/tree/master/picky-signtool), other crates do not depend on it and are pure Rust as far as I know (we use RustCrypto instead of ring for that reason).
There is no plan to rewrite the lief project in Rust just for picky-signtool.

RRRadicalEdward:
It would be possible if there is a Rust library for decoding, modifying, and encoding PE files.
The last time, I checked there wasn't any.

CBenoit:
> It would be possible if there is a Rust library for decoding, modifying, and encoding PE files. The last time, I checked there wasn't any.

I wonder if [goblin](https://crates.io/crates/goblin) or [object](https://crates.io/crates/object) could do the job?

RRRadicalEdward:
I just checked [object::write::pe::Writer](https://docs.rs/object/0.31.0/object/write/pe/struct.Writer.html) and it should do the job

CBenoit:
> I just checked [object::write::pe::Writer](https://docs.rs/object/0.31.0/object/write/pe/struct.Writer.html) and it should do the job

Great!

I renamed the issue to reflect how things could be actioned.

Note that picky-signtool is not a priority anymore for us (Devolutions), but pull requests from the community are welcomed :slightly_smiling_face:

PR #242, opened 2023-08-31
chore: bump NuGet package version + nuke picky-signtool

CBenoit:
One commit is bumping the NuGet package version, and the other one removes `picky-signtool`.

It was agreed with @awakecoding to completely remove `picky-signtool` from picky-rs repository at the moment:
- We don’t use it at Devolutions anymore (I don’t remember if we ever did actually),
- It increases the maintenance cost of the repository because it depends on [`lief-cwal`/`lief-cwal-sys`](https://github.com/Devolutions/lief-rs) that
  - is also depending on `picky`, causing a circular dependency, and
  - relies on a C++ library that must be build using `cmake` from a `build.rs`.
  - This would be addressed by https://github.com/Devolutions/picky-rs/issues/213, but we don’t have the bandwidth for that.

It is recommended to use [Jordan Borean’s OpenAuthenticode](https://github.com/jborean93/PowerShell-OpenAuthenticode) instead.

Closes https://github.com/Devolutions/picky-rs/issues/213


awakecoding (approved):
goodbye picky-signtool!

PR #261, opened 2024-04-02
feat: WASM bindings for SSH keys (and sk keys)

pacmancoder:
This PR adds new functionality to picky:
- SSH key generation and parsing API in WASM module - [ARC-208](https://devolutions.atlassian.net/browse/ARC-208)
- Updates C# bindings - adds missing `generate_ed25519` method for `PrivateKey` [ARC-62](https://devolutions.atlassian.net/browse/ARC-62)
- Adds support for `sk-ecdsa` and `sk-ed25519` key parsing (U2F-keys for use with security keys e.g. yubikey) to make picky functionality on par with RustCrypto's `ssh-key` crate - encode/decode for keys and certificates. (No signing or generation, as this would require additional logic to communicate with hardware U2F keys)

[ARC-208]: https://devolutions.atlassian.net/browse/ARC-208?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ARC-62]: https://devolutions.atlassian.net/browse/ARC-62?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pacmancoder:
NOTE: `ffi/dotnet` and `ffi/js` folders could be ignored during review, they are automatically generated

ffi/js/api/Argon2.d.ts
pacmancoder, line 1: Question: could we remove legacy JS bindings as we have a separate wasm module?

ffi/src/ssh.rs
```diff
@@ -149,6 +154,28 @@ pub mod ffi {
             Ok(Box::new(SshPrivateKey(key)))
         }
 
+        /// Generates a new SSH ed25519 Private Key.
+        ///
+        /// No passphrase is set if `passphrase` is empty.
+        ///
+        /// No comment is set if `comment` is empty.
+        pub fn generate_ed25519(passphrase: &str, comment: &str) -> Result<Box<SshPrivateKey>, Box<PickyError>> {
```
pacmancoder, line 162: Added missing ed25519 SSH key generation for C# bindings

ffi/wasm/src/ssh.rs
```diff
@@ -0,0 +1,99 @@
...
+    pub fn public_key(&self) -> SshPublicKey {
+        SshPublicKey(self.0.public_key().clone())
+    }
+}
+
+#[wasm_bindgen]
+pub struct SshPublicKey(pub(crate) picky::ssh::public_key::SshPublicKey);
+
+#[wasm_bindgen]
+impl SshPublicKey {
```
pacmancoder, line 73: Added ssh key generation, parsing, encoding as per [ARC-208](https://devolutions.atlassian.net/browse/ARC-208)

pacmancoder: Should we also add SSH certs in our API or it could be done as a follow-up?

pacmancoder:
NOTE: I reverted automatically generated bindings changes (dotnet & legacy js bindings) for ease of reviewing, I'll re-generate those bindings after code review will be finished

ffi/src/ssh.rs
pacmancoder: Question: could we remove legacy JS bindings as we have a separate wasm module?

pacmancoder:
For some reason, `jwe_ecdh_es_roundtrip` test becomes flaky, needs investigation before merge

ffi/src/ssh.rs
CBenoit: The ones generated by Diplomat, right? I would prefer if we could use Diplomat to generate both C# and JS, but JS required further investigation back then. We could delete it until we look further into it though.

ffi/wasm/src/ssh.rs
CBenoit: I’m fine with a follow-up! :+1:

CBenoit: AFAIK, Hub team really needed the API for managing SSH key pairs, but not the certs, so we can reevaluate later.

CBenoit:
> NOTE: I reverted automatically generated bindings changes (dotnet & legacy js bindings) for ease of reviewing, I'll re-generate those bindings after code review will be finished

Sure! :+1: 
Make sure to use the appropriate version of diplomat-tool. I think there is a just rule for all of this.

PS: I don’t remember if we update the diplomat JS bindings anymore since they don’t work as of now.

picky/src/ssh/private_key.rs
```diff
@@ -173,6 +205,17 @@ impl SshPrivateKey {
         &self.base_key
     }
 
+    pub fn inner_key(&self) -> Result<&PrivateKey, SshPrivateKeyError> {
+        match self.base_key() {
+            SshBasePrivateKey::Rsa(key) => Ok(key),
+            SshBasePrivateKey::Ec(key) => Ok(key),
+            SshBasePrivateKey::Ed(key) => Ok(key),
+            SshBasePrivateKey::SkEcdsaSha2NistP256 { .. } | SshBasePrivateKey::SkEd25519 { .. } => {
+                Err(SshPrivateKeyError::SkKeyDoesNotContainPrivateKeyData)
```
RRRadicalEdward, line 214: If a function can only fail in one case, it makes sense to return the value as `Option<...>` instead of `Result<..., ...>`

RRRadicalEdward (approved):
LGTM!  I only have one comment

picky/src/ssh/private_key.rs
pacmancoder: Agreed!

ffi/src/ssh.rs
pacmancoder: I'll remove generated files (`api` and `docs` directory) so it is guaranteed that we will not keep stale generated JS files in our repo source tree, and keep the manually written stuff, just to be able resume work on JS diplomat bindings when needed. Ideally, we could add generated files to git ignore (for C# too) and generate them only when needed (maybe add CI step?)

pacmancoder:
@CBenoit FYI - fixed flaky test for EC keys, we had an invalid `BigUint` -> fixed size EC field conversion, the leading zeros were missing, causing field size validation to fail sometimes

picky/src/ssh/private_key.rs
pacmancoder: Done

pacmancoder:
UPD: Diplomat bindings were updated

ffi/src/ssh.rs
CBenoit: For now we are checking the generated files on purpose, because:

- It can be read along with the main application code and reviewed (even briefly) to get a complete picture of how things work. (I always briefly check the generated code in PRs like this one.)
- I can see spot the differences in development environments (typically, the diplomat-tool version)
- The code is buildable and testable right after git clone and other tools like static analysis also work right away.

That being said, I think we should at least run a check to ensure that the generated code is not stale.

I’m also not against removing the generated code completely, but I don’t think it’s necessarily bad thing to have it either.

CBenoit (approved):
Good job! Thank you!

PR #262, opened 2024-04-05
feat(picky): add support for P521 NIST curve EC keys

pacmancoder:
`p512` supports curve arithmetic now, therefore the following P521 EC key functionality have been added:
- private/public generation
- signing/verifying
- SSH key generation, certificate signing
- jose algorithms
- updated FFI


CBenoit (approved):
LGTM! Thank you!!

PR #276, opened 2024-05-07
fix(picky): parse SSH public key comment til the end

CBenoit:
The parsing logic for the public key comments was stopping on the next whitespace instead, but comments may include multiple words separated by whitespaces.

Issue: ARC-217


pacmancoder (approved):
LGTM

CBenoit:
@pacmancoder It turns out the parsing logic for SSH certificates suffered the same issue so I also fixed that

PR #277, opened 2024-05-07
fix(putty): add `with_comment` method to public and private PuTTY keys

pacmancoder:
This PR adds `with_comment` methods to `PuttyPpk` and `PuttyPublicKey` structures.

Note that those structures were immutable by design, so `with_comment` will always return new instance with new comment and re-calculated MAC.


CBenoit (approved):
Since we need to recalculate the MAC anyway, this approach sounds reasonable :+1: 
LGTM!

PR #283, opened 2024-06-04
build(wasm): use vite to embed the WASM module in base64

irvingoujAtDevolution:
Use Vite to rebuild Picky WASM into Base64 encoded JavaScript. This approach have the following benefits for angular 18:
1. Import and use, no extra configuration
    In Angular 18, if we want to use WASM, we need to configure Vite (that is used bu Angular 18) to use [vite-plugin-wasm](https://www.npmjs.com/package/vite-plugin-wasm).
2. Zero interface change
3. Backward compatibility
    As it is just JavaScript file, any project can just import it without worrying about how to load WASM

### DIFF
--- a/ffi/wasm/README.md
+++ b/ffi/wasm/README.md
@@ -11,15 +11,14 @@ This should be run in the CI.
 2. Build the package: 
 
     ```
-    $ wasm-pack build --target web --scope devolutions --out-name picky
+    $ npm run build:wasm 
+    $ npm run build
     ```
 
-3. Rename `@devolutions/picky-wasm` to `@devolutions/picky` in `pkg/package.json`.
-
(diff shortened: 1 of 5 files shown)


CBenoit:
> In Angular 18, if we want to use WASM, we need to configure Vite (that is used bu Angular 18) to use [vite-plugin-wasm](https://www.npmjs.com/package/vite-plugin-wasm).

I understand that we only need to use this vite plugin when bundling our npm package together, but this is not required in the consuming project, am I correct?

irvingoujAtDevolution:
@CBenoit , yes, you are correct, consumers can use anything they want.

CBenoit (approved):
LGTM!

PR #287, opened 2024-06-25
chore(deps): update Cargo.lock

MathieuMorrissette:
Update cargo lock file to dismiss some dependabot alerts.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -361,16 +361,15 @@ dependencies = [
 
 [[package]]
 name = "curve25519-dalek"
-version = "4.1.1"
+version = "4.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


CBenoit:
Thank you!
However, this is breaking the CI. Could you try to target only the dependencies using `cargo update -p` and see if it’s enough?
Otherwise, we’ll have to update our MSRV requirements.

CBenoit (commented):
Thank you!
However, this is breaking the CI. Could you try to target only the dependencies using `cargo update -p` and see if it’s enough?
Otherwise, we’ll have to update our MSRV requirements.

MathieuMorrissette:
@CBenoit  done!

CBenoit (approved):
Nice! Thank you! 🙏

PR #317, opened 2024-10-29
fix(pkcs12): Added API to query and set MAC algorithm for PFX

pacmancoder:
- Add API to query/set MAC algorithm for PFX explicitly
- Added fallback logic to select SHA1 HMAC automatically if PBES1(3DES/RC2) is used


thenextman (approved):

CBenoit (approved):
LGTM

PR #319, opened 2024-10-30
fix(ffi): remove Pkcs12Mac iterations from FFI interface

thenextman:
The `Iterations` getter on Pkcs12MacAlgorithm doesn't build in dotnet:

> error CS0266: Cannot implicitly convert type 'uint*' to 'uint?'. An explicit conversion exists (are you missing a cast?)

I'm unsure right now how to fix this in diplomat, but the field seems to just be for completeness / an implementation detail. So for now, I remove it from the FFI interface.

### DIFF
--- a/ffi/src/pkcs12.rs
+++ b/ffi/src/pkcs12.rs
@@ -78,10 +78,6 @@ pub mod ffi {
         pub fn hash_algorithm(&self) -> Pkcs12HashAlgorithm {
             self.0.hash_algorithm().into()
         }
-
-        pub fn iterations(&self) -> Option<Box<u32>> {
-            self.0.iterations().map(Box::new)
-        }
     }
 
     #[diplomat::opaque]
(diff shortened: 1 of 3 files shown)


ffi/src/pkcs12.rs
```diff
@@ -78,10 +78,6 @@ pub mod ffi {
         pub fn hash_algorithm(&self) -> Pkcs12HashAlgorithm {
             self.0.hash_algorithm().into()
         }
-
-        pub fn iterations(&self) -> Option<Box<u32>> {
```
CBenoit, line 82: Can you try with `Option<u32>` instead? I didn’t thought twice when I reviewed this, but it’s a bit strange to returned a boxed u32.

pacmancoder: If I remember correctly, diplomat didn't allow returning unboxed `Option<u32>`  😢

thenextman: Seems right

>diplomat-tool dotnet ./dotnet/Devolutions.Picky/Generated/ -l ./dotnet-interop-conf.toml
>A non-reference type was found inside an Option<T>: Option<u32>

CBenoit (approved):
LGTM!

Issue #325, opened 2024-11-22
picky-asn1-x509 0.14 release has few build issues


elmarco:
```
error: couldn't read `src/../../test_assets/missing_authority_key_identifier_field.crt`: No such file or directory (os error 2)
   --> src/certificate.rs:306:23
    |
306 |         let encoded = include_bytes!("../../test_assets/missing_authority_key_identifier_field.crt");
    |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

And a couple of warnings:
```
warning: this function depends on never type fallback being `()`
   --> src/algorithm_identifier.rs:301:13
    |
301 | /             fn visit_seq<A>(self, mut seq: A) -> Result<Self::Value, A::Error>
302 | |             where
303 | |                 A: de::SeqAccess<'de>,
    | |______________________________________^
    |
    = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
    = note: for more information, see issue #123748 <https://github.com/rust-lang/rust/issues/123748>
    = help: specify the types explicitly
note: in edition 2024, the requirement `!: Deserialize<'de>` will fail
   --> src/macros.rs:14:14
    |
14  |         $seq.next_element()?.ok_or_else(|| {
    |              ^^^^^^^^^^^^
    |
   ::: src/algorithm_identifier.rs:359:25
    |
359 |                         seq_next_element!(seq, AlgorithmIdentifier, "sha algorithm identifier");
    |                         ----------------------------------------------------------------------- in this macro invocation
    = note: `#[warn(dependency_on_unit_never_type_fallback)]` on by default
    = note: this warning originates in the macro `seq_next_element` (in Nightly builds, run with -Z macro-backtrace for more info)

warning: creating a shared reference to mutable static is discouraged
   --> src/oids.rs:33:36
    |
33  |                   if let Some(oid) = &OID { oid.clone() } else { unreachable() }
    |                                      ^^^^ shared reference to mutable static
...
42  | / define_oid! {
43  | |     // x9-57
44  | |     DSA_WITH_SHA1 => dsa_with_sha1 => "1.2.840.10040.4.3",
45  | |     // x9-42
...   |
275 | |     LOCAL_PROTECTION_DESCRIPTOR => local_protection_descriptor => "1.3.6.1.4.1.311.74.1.8",
276 | | }
[...]

Issue #324, opened 2024-11-22
picky-asn1 0.9 release is missing dev-dependencies 


elmarco:
For some reason, it's missing picky-asn1-der, and failing to build doctest. Maybe you need workspace for that to work?

```
   Doc-tests picky_asn1
     Running `/usr/bin/rustdoc --edition=2021 --crate-type lib --color auto --crate-name picky_asn1 --test src/lib.rs --test-run-directory /builddir/build/BUILD/rust-picky-asn1-0.9.0-build/picky-asn1-0.9.0 -L dependency=/builddir/build/BUILD/rust-picky-asn1-0.9.0-build/picky-asn1-0.9.0/target/rpm/deps -L dependency=/builddir/build/BUILD/rust-picky-asn1-0.9.0-build/picky-asn1-0.9.0/target/rpm/deps --extern oid=/builddir/build/BUILD/rust-picky-asn1-0.9.0-build/picky-asn1-0.9.0/target/rpm/deps/liboid-ebff3ae5c9007caf.rlib --extern picky_asn1=/builddir/build/BUILD/rust-picky-asn1-0.9.0-build/picky-asn1-0.9.0/target/rpm/deps/libpicky_asn1-82c7b3838a677f6c.rlib --extern serde=/builddir/build/BUILD/rust-picky-asn1-0.9.0-build/picky-asn1-0.9.0/target/rpm/deps/libserde-ad2146265121d7c5.rlib --extern serde_bytes=/builddir/build/BUILD/rust-picky-asn1-0.9.0-build/picky-asn1-0.9.0/target/rpm/deps/libserde_bytes-1ddbdc6dc57c90ba.rlib -C embed-bitcode=no --check-cfg 'cfg(docsrs)' --check-cfg 'cfg(feature, values("chrono", "chrono_conversion", "time", "time_conversion", "zeroize"))' --error-format human`

running 13 tests
test src/wrapper.rs - wrapper::HeaderOnly (line 452) ... FAILED
test src/tag.rs - tag::TagPeeker (line 201) ... FAILED
test src/wrapper.rs - wrapper::BitStringAsn1Container (line 555) ... FAILED
test src/wrapper.rs - wrapper::OctetStringAsn1Container (line 607) ... FAILED
test src/wrapper.rs - wrapper::Optional (line 657) ... FAILED
test src/bit_string.rs - bit_string::BitString::is_set (line 145) ... ok
test src/bit_string.rs - bit_string::BitString (line 21) ... ok
[...]

PR #327, opened 2024-11-26
Add picky-test-data crate (and various fixes)

elmarco:
Fixes for #324 & #325

/!\ The OnceLock path requires bumping msrv to 1.70.


CBenoit (commented):
I think bumping MSVC to 1.70 (more than 1 year ago) is acceptable :+1: 

Also, you don’t have to care too much about picky-server, as it’s now unmaintained. I was considering removing it from the repository soon to avoid confusions.

Before we merge, do you think you could update the CHANGELOG files of the relevant crates? (e.g.: https://github.com/Devolutions/picky-rs/blob/master/picky-asn1-x509/CHANGELOG.md)

picky/src/ssh/certificate.rs
```diff
@@ -830,7 +830,7 @@ impl SshCertificateBuilder {
 pub mod tests {
     use super::*;
     use crate::ssh::private_key::SshPrivateKey;
-    use crate::test_files;
+    use picky_test_data;
```
CBenoit, line 833: (This remained at many places.)

CBenoit (approved):
LGTM!

PR #330, opened 2024-12-06
chore: add copy-license-files.ps1 script

awakecoding:
- Remove copies of LICENSE-MIT and LICENSE-APACHE files
- Add copy-license-files.ps1 script to copy it in all crates
- Updated .gitignore to ignore license file copies

copy-license-files.ps1 should be run prior to publishing crates


decathorpe:
I'm pretty sure this doesn't work as expected - picky-asn1-x509 v0.24.2 which seems to have been published since this was pushed doesn't contain license files: https://diff.rs/picky-asn1-x509/0.14.1/picky-asn1-x509/0.14.2/Cargo.toml

Was that script not run? Or does `cargo package` ignore files that are listed in `.gitignore`? I only found [this issue](https://github.com/rust-lang/cargo/issues/14411) which should be fixed in latest stable rust.

CBenoit:
Hi @decathorpe 

Indeed, I reevaluated this solution and here is my conclusion:

We should use symlinks instead of using this script for copying the files. The reason is that cargo publish will not let us publish the crates if the repository is in a "dirty" state. We get into a dirty state when copying files. We can bypass this check, but then cargo will stop packaging the vsc information into the crate, and that’s not good either.

Today, the symlinks should be fine, since we’re going to let the CI do the publishing work for us since https://github.com/Devolutions/picky-rs/pull/337 is merged, and the environment of the developers will not matter anymore.

However, I forgot to add back the symlinks and to remove the script before merging [the release PR](https://github.com/Devolutions/picky-rs/pull/338) and so they are missing.

We need to open a PR to add the symlinks again.

decathorpe:
Thank you for taking a look!

PR #334, opened 2025-01-08
picky-asn1-x509: pkcs7: implement `ContentInfo`

TheBestTvarynka:
Hi,
In this PR, I've implemented the `ContentInfo` structure. We need it for correct DPAPI structure serialization/deserialization. The existing `EncapsulatedContentInfo` is different and is not suitable.


CBenoit (approved):
LGTM!

PR #339, opened 2025-01-16
fix: symlinks to license files in packages

CBenoit:
Use symlinks instead of copying files to avoid a “dirty” state during cargo publish and preserve VCS info. With #337 merged, CI handles publishing consistently, so developer environments no longer matter.


CBenoit:
Addressing https://github.com/Devolutions/picky-rs/pull/330#issuecomment-2594086532
cc @decathorpe

awakecoding (approved):

PR #342, opened 2025-02-04
feat(picky-krb): implement Kerberos encryption without a checksum

TheBestTvarynka:
Hi,
I added the possibility of Kerberos encryption but without a checksum. We need this functionality to support `SECBUFFER_READONLY` and `SECBUFFER_READONLY_WITH_CHECKSUM` flags for security buffers in `sspi-rs`. Related to: https://github.com/Devolutions/sspi-rs/issues/120 

Additionally, I reviewed the `crypto` module in `picky-krb`. It seems like Kerberos encryption algorithms were implemented a long time ago. We can refactor it by using [`cts`](https://docs.rs/cts/latest/cts/) and [`aes`](https://docs.rs/aes-gcm/latest/aes_gcm/) crates and get rid of the custom `cts` mode implementation. Let's create an issue for the future.

Docs & reference:

* [`SecBuffer` structure (sspi.h)](https://learn.microsoft.com/en-us/windows/win32/api/sspi/ns-sspi-secbuffer).

### DIFF
--- a/picky-krb/src/crypto/aes/aes128_cts_hmac_sha1_96.rs
+++ b/picky-krb/src/crypto/aes/aes128_cts_hmac_sha1_96.rs
@@ -1,10 +1,13 @@
 use rand::rngs::OsRng;
 use rand::Rng;
 
-use crate::crypto::{Cipher, CipherSuite, KerberosCryptoError, KerberosCryptoResult};
+use crate::crypto::{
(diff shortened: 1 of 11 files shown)


CBenoit (approved):
Nice! LGTM.

> Additionally, I reviewed the `crypto` module in `picky-krb`. It seems like Kerberos encryption algorithms were implemented a long time ago. We can refactor it by using [`cts`](https://docs.rs/cts/latest/cts/) and [`a`](https://docs.rs/aes-gcm/latest/aes_gcm/)

Good catch. Refactoring out this code is much welcomed! :+1:

PR #351, opened 2025-03-01
fix(picky)!: comply with draft-cavage-http-signatures-12 when signing HTTP messages

sp1ff:
Not sure if picky is attempting to be compliant with "draft-cavage-http-signatures-12" (i.e. the algorithm used throughout the Fediverse), or not. If so, the request target field in the signing string gets its own "pseudo-header name": `(request-target)`. This PR changes that & updates the test suite accordingly.

Commit message follows:

For reasons unclear to me, the `http-signature` module treats the request target specially in that it is added to & read from the signing string with no header name (unlike, say, `(created)` or `(expires)`). As we see [here](https://datatracker.ietf.org/doc/html/draft-cavage-http-signatures-12#page-9), it, too,  has a "psuedo-header" name: `(request-target)`.

It's absence from the signing string was breaking my attempts to use picky to validate message signatures from Mastodon.

### DIFF
--- a/picky/src/http/http_signature.rs
+++ b/picky/src/http/http_signature.rs
@@ -555,6 +555,9 @@ impl<'a> HttpSignatureBuilder<'a> {
                         let key = split.next().expect("there is always at least one element in the split");
                         if let Some(value) = split.next() {
                             match key {
+                                Header::REQUEST_TARGET_STR => {
+                                    headers.push(Header::RequestTarget);
+                                }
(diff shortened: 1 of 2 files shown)


CBenoit (approved):
Thank you! It is a you said. Probably something was misunderstood at the time. Let’s merge this compliant version of the code. We’ll advertise this as a breaking change for the next major release.

PR #373, opened 2025-06-04
fix(picky-krb): typo in `AuthenticatorInner::authenticator_vno`

TheBestTvarynka:
Hi,
I noticed that we have a typo in the `AuthenticatorInner` field name.


CBenoit (approved):
LGTM!

PR #381, opened 2025-06-23
fix(picky-krb): `EncTicketPart` structure;

TheBestTvarynka:
Hi,
During dev testing and debugging I noticed that the `sspi-rs` cannot parse the decrypted ticket data generated by the KDC. After that, I discovered that the `EncTicketPart` structure is defined incorrectly in `picky-krb`.

### DIFF
--- a/picky-krb/src/data_types.rs
+++ b/picky-krb/src/data_types.rs
@@ -14,7 +14,9 @@ use std::io::{Read, Write};
 use std::marker::PhantomData;
 use std::{fmt, io};
 
-use crate::constants::types::{AUTHENTICATOR_TYPE, ENC_AP_REP_PART_TYPE, KRB_PRIV_ENC_PART, TICKET_TYPE};
+use crate::constants::types::{
+    AUTHENTICATOR_TYPE, ENC_AP_REP_PART_TYPE, ENC_TICKET_PART_TYPE, KRB_PRIV_ENC_PART, TICKET_TYPE,
+};
 use crate::messages::KrbError;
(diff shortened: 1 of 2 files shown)


picky-krb/src/data_types.rs
```diff
@@ -214,6 +216,8 @@ pub struct EncTicketPart {
     pub authorization_data: Optional<Option<ExplicitContextTag10<AuthorizationData>>>,
 }
 
+pub type EncTicketPart = ApplicationTag<EncTicketPartInner, ENC_TICKET_PART_TYPE>;
```
CBenoit, line 219: issue? The ApplicationTag only relevant in the wrapping structure, so it seems to me that the fix would be to add the ApplicationTag in the wrapping structure directly.

TheBestTvarynka: > so it seems to me that the fix would be to add the ApplicationTag in the wrapping structure directly.

What do you mean? 😅 

> The ApplicationTag only meaningful in the context of SEQUENCE

yes, and the `EncTicketPartInner` is a SEQUENCE...

CBenoit: Context and application tags are only meaningful in the context of another data structure, to distinguish between optional fields. That's why it does not seems appropriate to me to hardcode the tag into the type of the field directly. Maybe in this case it's used at only one place, but the same data structure could be wrapped into a different tag depending on the wrapping structure, so it's appropriate to specify the tag in the enclosing structure.

PR #397, opened 2025-09-05
Improve support for SHA3 digests within EC and RSA certificates

c0d3d:


c0d3d:
@CBenoit sorry - formatting didn't pass before. I've fixed it now

PR #400, opened 2025-09-09
build: optimize binary size for FFI builds

CBenoit:
Aim for good balance between size and performance.

- Baseline: 7.4M
- strip = "symbols": 6.0M
- lto = true: 4.7M
- opt-level = "z": 3.5M
- codegen-units = 1: 3.3M
- panic = "abort": 3.0M
- opt-level = 3 for crypto libs: 3.1M


CBenoit:
Note: in the last few years, we never had to deal with panics, so I went ahead with `panic = "abort"`.

PR #405, opened 2025-09-17
chore: upgrade to 2024 Rust edition

RRRadicalEdward:


ffi/src/pem.rs
```diff
@@ -67,12 +67,14 @@ pub mod ffi {
...
-#[no_mangle]
+/// * `len` mustn't point to a NULL-pointer.
+/// * returned data should not be modified.
+#[unsafe(no_mangle)]
 pub unsafe extern "C" fn Pem_peek_data(pem: Option<&ffi::Pem>, len: *mut usize) -> *const u8 {
     if let Some(pem) = pem {
         let data = pem.0.data();
-        *len = data.len();
+        // Safety: `len` is not a NULL-pointer.
+        unsafe { *len = data.len() };
```
CBenoit, line 77: suggestion: This is a little bit out of scope, but while at it, I would suggest this version:

```rust
/// Returns a borrowed pointer to the bytes contained in this `Pem`.
///
/// On success, writes the length of the data to `len` and returns a non-null
/// pointer to the first byte. If `pem` is `None`, writes `0` to `len` and
/// returns a null pointer.
///
/// # Safety
///
/// * `len` must be a valid, non-null, properly aligned pointer to writable
///   memory for a `usize`. This function will write to `*len`.
/// * If `pem` is non-NULL, it must reference a live, valid `Pem`.
/// * The returned pointer is **borrowed**. It is only valid for as long as
///   the referenced `pem` remains alive and is not mutated in a way that could
///   reallocate or invalidate its internal buffer.
/// * The returned memory must be treated as read-only. Do not write to it and
///   do not free it. The bytes are not NUL-terminated; use the length written
///   to `len`.
#[no_mangle]
pub unsafe extern "C" fn Pem_peek_data(pem: Option<&ffi::Pem>, len: *mut usize) -> *const u8 {
    if let Some(pem) = pem {
        let data = pem.0.data();
        // SAFETY: caller guarantees `len` is non-null, properly aligned, and writable.
        unsafe { len.write(data.len()) };
        data.as_ptr()
    } else {
        // SAFETY: caller guarantees `len` is non-null, properly aligned, and writable.
        unsafe { len.write(0) };
        core::ptr::null()
    }
}
```

CBenoit:
LGTM!

> CI fails because I increased MSRV to 1.85 (the 2024 edition release version). We will need to update CI checks to expect MSRV to be 1.85 in this PR as well. I will do it once the current changes are approved.

Please, you can go ahead. Thank you!

ffi/src/pem.rs
RRRadicalEdward: Is there a difference between `unsafe { *len = data.len() };` and `unsafe { len.write(data.len()) };` ?
In the first case, we deference the `len` pointer, but it should also happen in the case of `len.write(...)`, shouldn't it? Asking because I'm curious 😄.

CBenoit: That’s a very good question! `*len` first dereferences the raw pointer, yielding an usize lvalue, and then the assignment overwrites the pointed-to memory. It actually _reads_ the memory first. The value is also dropped. `len.write(…)` performs the same write into the pointer’s target memory, but crucially: it does so without ever trying to read or drop the old value in memory. So overall: `*len` is more dangerous because:
- Reading uninitialized memory is an immediate UB
- _If_ `Drop` is implemented, we may ending double-freeing, or creating other undesirables side effects.
Not to say that `*ptr = value` is always bad, but when there is no need to read the value and to drop it, it looks much safer to use `write` :slightly_smiling_face: 
(Of course, I’m probably the culprit of this code :joy:)

CBenoit: Side note: writing 0 to pointer is also a way of ensuring the memory is initialized even if it failed, as a defensive way of avoiding UBs.

.github/workflows/ci.yml
```diff
@@ -104,11 +104,11 @@ jobs:
...
           - crate: picky-asn1
-            msvc: "1.68"
+            msvc: "1.85"
           - crate: picky-asn1-der
-            msvc: "1.68"
+            msvc: "1.85"
           - crate: picky-asn1-x509
-            msvc: "1.70"
+            msvc: "1.85"
           - crate: picky
```
CBenoit, line 112: issue: I think you also need to update the README.md / CHANGELOG.md to reflect this

PR #517, opened 2026-07-21
ci: run the .NET binding tests

thenextman:
Follow-up to #516. Adds a CI job that builds the native library and runs the .NET binding tests (`Devolutions.Picky.Tests`) on Windows — the suite that would have caught the ABI mismatch fixed in #516, but which no workflow was running.

## What the job does
- Builds the native library in release (`cargo build --release -p picky-ffi`) — matches what ships and avoids debug-only std precondition checks at the FFI boundary.
- Installs the iOS workload (required to build the multi-target `Devolutions.Picky` project, same as `nuget.yml`'s `build-managed`).
- Builds the tests against the **committed** bindings (`-p:GeneratePackageOnBuild=false`), so a stale/mismatched binding fails CI.
- Places the native library next to the test host (the project sets `CopyToOutputDirectory=Never`), provisions the fixtures, and runs `dotnet test`.
- Wired into the `success` gate.

## Also included
`X509Tests.X509Certificate2Conversion` looked up the Subject Key Identifier extension by the friendly-name string `"subjectKeyIdentifier"`, which newer .NET runtimes don't match (the canonical name is `Subject Key Identifier`). Switched to the OID `2.5.29.14` so the lookup is runtime-independent.

Verified locally: 26/26 tests pass.

## Follow-ups (not in this PR)
- The tests read fixtures via brittle relative paths that resolve to the repo's parent; the job provisions fixtures there to match. Worth repointing them at `picky-test-data/test_assets`.
- A debug native build trips a std `from_raw_parts` precondition (empty slice passed as a null pointer at the FFI boundary); harmless in release but a real latent diplomat-FFI quirk.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -123,10 +123,55 @@ jobs:
       - name: cargo check ${{ matrix.crate }}
         run: cargo +${{ matrix.msvc }} check -p ${{ matrix.crate }}
 
+  dotnet-tests:
+    name: .NET binding tests
+    needs: [formatting]
+    runs-on: windows-latest
+
+    steps:
+      - uses: actions/checkout@v6
+
+      # Release matches what ships (the ffi-production/NuGet build) and avoids debug-only
+      # std precondition checks at the FFI boundary.
+      - name: Build native library
+        run: cargo build --locked --release -p picky-ffi
+        shell: pwsh
+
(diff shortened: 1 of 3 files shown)


CBenoit (approved):
LGTM! I pushed some fixes, it seems to be green now.

